PR #10, opened 2014-01-27
WIP: symplectic solver based on velocity verlet

acroy:
This PR adds two functions, `verlet_fixed` and `verlet_hh2`, which implement the velocity Verlet method with fixed and adaptive time-steps, respectively. The latter uses an estimate of the error based on comparing a full step with two half-steps. This function is exported as `verlet`.

The interface is largely compatible with the proposal in #7. The main difference is, that "velocity" and "position" are
specified (and returned) separately, instead of having one solution vector. Moreover, the acceleration (or force) currently only depends on time and position.

There are still some other things to do (absolute and relative tolerances, intermediate steps in `tspan`, ...), which partly depend on the outcome of #7.

Examples can be found here: [harmonic oscillator](https://gist.github.com/acroy/8646927) and [pendulum](https://gist.github.com/acroy/8647126)

Comments and suggestions are very welcome.

### DIFF
--- a/src/symplectic.jl
+++ b/src/symplectic.jl
@@ -50,23 +50,22 @@ end
 ##
 ## The function takes as a first argument a function, a(t,x),
 ## of time and position (the acceleration). The second argument
-## specifies a list of times for which a solution is requested. The last
+## specifies an intial time t_0 and a final time tf for which a solution is requested. The last
 ## two arguments are the initial conditions x_0 and v_0.
 ##
 ## Optional keywords are atol (error tolerance), norm (function to calculate the error)
 ## and points=:all|:specified (flag to indicate type of output).
 ##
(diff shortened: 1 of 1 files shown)


mauro3:
Another reference is [Hairer,  Wanner & Lubich, 2002](http://link.springer.com/book/10.1007/978-3-662-05018-7), chapter VIII.

mauro3:
Is there actually a reason this was never merged?

acroy:
Good question ;-) I guess there might be two reasons: 1) Although the
method works it is very crude IMO and 2) We never settled the API - for
example, the current version doesn't use "the right-hand side" of the ODEs,
but rather the force...

mauro3:
Thanks for the clarification.  Leave open or close?

Issue #25, opened 2014-03-05
Possibly implement dopri5


jtravs:
Just a suggestion, but as discussed here: https://github.com/JuliaLang/julia/issues/75#issuecomment-36766329
One way around the licensing issues with the current ODE codes is to derive codes from the solvers at:
http://www.unige.ch/~hairer/software.html

In particular the MATLAB ode45 is very similar to the dopri5 code at:
http://www.unige.ch/~hairer/prog/nonstiff/dopri5.f
and this is the same set of coefficients used in ODE.jl (the Dormand and Prince pair).

These solvers are BSD licensed, are of very high quality, are theoretically explained in detail in the books by the authors:
Hairer, Norsett and Wanner (1993): Solving Ordinary Differential Equations. Nonstiff Problems. 2nd edition. Springer Series in Comput. Math., vol. 8.

What is particularly nice about dopri5 is the dense output. i.e. if the ODE solver takes a natural step from t0 to t1 based on the error control, dense output is efficiently available anywhere between t0 and t1.

Issue #11, opened 2014-02-02
Event functions


tomasaschan:
For some problems, it is interesting to run the solver until something happens - specified by (part of) the solution at that point. In those cases, it can be hard to know on beforehand exactly when this happens, so it's difficult to give an upper limit to the solution domain.

For example, think about a program modelling a bouncing ball: it's just solving F = ma for a free-falling object, except every time the ball hits the floor, we want to change sign of the y-component of its velocity. Without event functions, one has to solve for one parabola at a time, obtain the final components of the velocity, change the sign and use as initial values for the next - with them, one can let the solver do the work.

It could also be nice to be able to stop the simulation when some condition is met, rather than after a final time - for a system that you expect to stabilize but you don't know how long it's going to take, you can then stop when the rate of change in the solution is small enough.

Basically, the solvers should support some way of saying `every time X happens, do Y` where `Y` can be anything from changing some part of the problem to stopping the solver. As noted in #7 it could also be interesting to support only emitting solution points when some condition is met, which could also be specified by an event function.

What is a good way of implementing this in the current solvers? What should the user interface look like?

spencerlyon2:
Any update on this? Does anyone have a working example of using callbacks to emulate the events functionality of Matlab?

mauro3:
I don't think so. The new RK solvers have dense output (although only 3rd order polynomials).  This could then be used to detect events.

spencerlyon2:
I needed this, so I implemented a routine that uses any integrator in ODE.jl. The routine is in [this gist](https://gist.github.com/spencerlyon2/d64b4e95409b1ef95295)

I'm quite positive this is one of the most inefficient ways of doing it, but it works for my needs right now and I thought I'd share.

acroy:
Since it just came up in #33: At some point I suggested a simple way to have a [progress callback for Sundials](https://github.com/JuliaLang/Sundials.jl/issues/40#issuecomment-72334904), which might in principle also work for our purposes?

geandersoncarvalho:
I am new in programming language specially jula. spencerlyon2 if I use your routine, how can I implemented this to solve a ODE with condition stop? I mean what can I have to do? Put a file in ODE folder with your routine? I don't now.

geandersoncarvalho:
I am new in julia programming language.
I am trying to solve an ODE but I need to set a stop condition. spencerlyon2 I try to use your routine as a tentative to set the this, but when I ran your routine even with your example I get the following error messsage
[...]

PR #33, opened 2014-03-20
[WIP] Port of DOP853

helgee:
This is an almost complete port of DOP853 (the dense output functionality is still missing) with quite good performance (https://groups.google.com/forum/#!searchin/julia-users/dop853/julia-users/41B3qULjF-0/WqGa-Lc04AkJ), approximately 2-3x slower than Fortran (without compiler optimizations) for longer runs. Since the 5-th order version DOPRI5 looks rather similar implementing it as well should not be a huge amount of work

See also: #25

To Do:
* [ ] API adjustment
* [ ] Dense output
* [X] DOPRI5
* [X] Write tests
* [ ] Documentation

* What else?
* Is it a good idea to separate different solver families in different source files?


helgee:
I apologize for letting this PR go stale. Should I continue with this or would it make more sense to just add the 8th-order Dormand-Prince coefficients to the new Runge-Kutta solvers?

In that case I would still need facilities for event detection (e.g. a user function that gets called after every integration step and dense output) for ODE.jl to be useful to me. But until that is available I will use my wrapper for the Fortran codes (https://github.com/helgee/Dopri.jl).

tlycken:
Adding the 8th-order coefficients would definitely be a nice contribution regardless of other things.

I personally hope that event detection can be handled with the approach outlined in #49, with support across all solvers in ODE.jl, but that's quite a large undertaking. Also, I don't have the opportunity to use ODE.jl regularly anymore - and thus my ability to invest time in it is also limited - so my opinion on the matter carries very little weight nowadays :)

acroy:
I think we should stick to the current implementation for anything related to (explicit) RK methods. If you could add the 8th order coefficients this would be great.

If you have ideas or needs for event functions, please post them in #11.

helgee:
Alright. I will close this then.

mauro3:
It is easy to add the method to the existing RK solvers, you just need to provide a tableau like:
https://github.com/JuliaLang/ODE.jl/blob/01620ad46321b2088f4b90be1099b903c43dd788/src/runge_kutta.jl#L135

At the moment dense output is only 3rd order. To change that would require some code changes.  I hope to implement a general scheme at some point but cannot promise a time frame.  Same goes for event location.

PR #44, opened 2014-06-25
dense output API

berceanu:
use matlab-like `refine` parameter


ChrisRackauckas:
I'm going to close this since it looks like there is no actual code coming to back up this change. But discussing how to improve dense output is something that would be good for an issue.

Issue #40, opened 2014-06-15
dense output support


berceanu:
I would like to know if dense output, in particular for ode45, is in the works or not. 
If not, I could give it a try, however 
1. i do not know if the current codebase will not change completely due to the GPL licensing (what are the plans?)
2. i do not know if dense output can be integrated nicely into the existing routines, or a rewrite is anyhow in order

Issue #27, opened 2014-03-11
Iterator versions of ODE solvers


ivarne:
I was thinking about https://github.com/JuliaLang/ODE.jl/issues/11 and just relaised that an iterator version of the ODE solvers might be a viable solution. As this is a different idea than the event function, I opened a new issue.

``` julia
ode = ode_45_iter(f, y0, [0,Inf])
for t, y in ode
    if y[1] > 10
        break
    end
end
# and if we want the ODE_iter to store the intermediary solutions
# (maybe that should be left as a task for the user).
t_out, y_out = get_ode_solution_vectors(ode)
```

This can also be very convenient for plotting partial results and to be able to track the solvers progress.

PR #49, opened 2014-10-21
[WIP] Adding iterators

pwl:
Hi,
I have been playing around with iterator version of `oderkf` and came up with my own implementation.  My idea was what follows:  make `oderkf(F,x0,tspan::Vector)` behave as previously and add `oderkf(F,x0,tstart::Number)` which creates an iterator.  I reimplemented `oderkf(F,x0,tspan::Vector)` using the iterator version of `oderkf`, so that both implementations are not redundant.  I also made changes to the keyword arguments accepted by `oderkf`
- New keyword arguments: `minstep`, `maxstep`, `initstep`, I decided that `hmin` and `hmax` from the previous implementation were already arbitrary, so why not allow the user to define his own problem specific values?  In the previous implementation `hmin` and `hmax` were computed from maximal and minimal time, for iterator this is impossible because, in principle, we might not now the maximal time.  Therefore, I set the default value of `minstep` to `eps(T)^(1/3)`, where `T` is the type of `tstart`; the default value for `initstep` is `minstep` while `maxstep=1/minstep`.  Alternatively we could set `maxstep=Inf`, but all these values are a matter for a discussion.
- I added a `tstop=Inf` optional argument to the iterator version of `oderkf`, naturally `oderkf` stops the integration exactly when `t=tstop` is reached.
- I changed the reaction for `abs(h)<abs(hmin)`, now it produces a warining and does not output `x` (only `t` and `h`).  Printing `x` is rather annoying when `x` is a large vector (the value of `x` is the last step before `abs(h)<abs(hmin)` so it is returned by the solver anyway).
- With `tstop=Inf` it is possible to integrate the system ad infinitum, so I had to add a test for finiteness of the resuts.

The implementation of the non-iterator version of `oderkf` is simply

```
function oderkf(F,x0,tspan::Vector,coefs...;args...)
    t0    = tspan[1]
    tstop = tspan[end]
    tout = Array(typeof(t0),0)
    xout = Array(typeof(x0),0)
    for (t,x) in oderkf(F,x0,t0,coefs...;args...,tstop=tstop)
        push!(tout,t)
        push!(xout,x)
    end
    return (tout,xout)
end
```

You can also use `collect(ode45((t,x)->x, 1., 0., tstop = 10))`.

You can now do something similar to what @ivarne had in mind in issue #27 :

```
ode = ode45((t,x)->x, 1., 0.)
for (t,y) in ode
    if y > 10
        @show (t,y)
        break
    end
end
```

With iterators we would be one step closer to implementing root finding for solutions to ODE's.  I am curious if iterators could get along with dense output #40 #44 .

PS I know that API is not fixed yet, but I needed iterators for my own purposes and decided to share this draft with you.

Cheers,
Paweł


pwl:
This seems to be a hot season for PRs, any thoughts on iterator versions of solvers?   I am willing to rebase this PR to the current master if anyone is interested in merging it.

tlycken:
I love this API idea, so I say do it!

In the interest of keeping the API consistent, we probably want to make this style possible in other solvers, but if that seems like a daunting task there's no problem in splitting that into more PR's. Also, note that @acroy implemented the keywords for `minstep` etc (including a nice estimator for the initial step) in #59 so you might want to base on that branch to avoid conflicts.

tlycken:
This would probably also be interesting for creating a better solution to timholy/ProgressMeter.jl#15, so there are real use cases.

pwl:
Ok, I think I will wait until #59 is merged and rework this PR after that.

tlycken:
Now that both #59 and #61 are merged, this is probably a good time for this too, if you're up for it :)

acroy:
OTOH it might make sense to wait for @mauro3's RK implementation. Otherwise the iterator version will also count as "derived" from `oderkf`.

tlycken:
Good point.

pwl:
I agree with @acroy, I will wait and build on top of @mauro3's work.

mauro3:
The pressure is on...

berceanu:
OK, so I want to help build the iterator version, because I really need it at this point.
Are the any outstanding issues with @mauro3's PR or can we merge that and move on? :)

mauro3:
From my side, it's ready to be merged.

An alternative to your problem in #71 could also be to introduce a `poststep` function which gets executed after each step.  This could be useful for diagnostics and such as well. Less general than iterators but probably a lot easier to program.

berceanu:
It's not just for the normalization problem that I need it. It is also useful for visualizing the convergence, so that I can see for example when the steady state is reached.

pwl:
If @mauro3 PR is merged in time I will be able to take a look at the iterators this weekend and rework them to fit the new solvers.

berceanu:
That would be great! Please, let me know how I can be of assistance.

acroy:
I guess `poststep` is a simple example of an event function and would be very useful. The question is how to realize the callback?

berceanu:
@pwl any progress on implementing the iterator version on top of @mauro3 's new RK code? :)

pwl:
@berceanu Sorry for not keeping you up to date.  I keep this issue at the back of my head constantly but the last week was pretty busy and currently I am at a workshop (which lasts until Friday).  I gave the code a look some time ago and implementing iterators will take considerably more time than I anticipated.  The reason is I think it is necessary to unify fixed step and adaptive methods to avoid wasting resources into supporting two separate iterators, and this would be a major rework of the existing source code.  From what portions of code I understand it seems possible, but as I said, it might take me some time to dig into the algorithms.

If you are in a hurry and you need the iterators right now feel free to use my fork, I have been using this implementation for quite some time and it works flawlessly (although some new functionality like points=:specified or custom types are not implemented there).  The API is described at the beginning of this PR.  Alternatively you could give this issue a shot yourself, if you have more time than I do.

mauro3:
One thing I noted yesterday using DASSL is that the error message for an error in the objective function did not report in which function the error happened and was thus pretty much useless.  I suspect because of the task-based iterator.  Is there a way around this?  Or is there at least an issue in Julia tracking this?

pwl:
Certainly, that's caused by the coroutines implementation of iterators in DASSL.  I don't know if there is any way to fix this, I started a "discussion" on a Google groups [1] but there wasn't much interest in it.  I chose to implement iterators via coroutines for DASSL because it was the least invasive method (essentially I just added one `produce()` statement to the main function) with practically no downsides, apart from the error reporting issues.  So coroutines provide a sort of poor man's iterators.

I think that for ODE we should have a proper iterator implemented via specialized `start`, `next` and `stop` routines, that's why it might take a little bit longer to get it right.

[1] https://groups.google.com/forum/#!topic/julia-users/R9yILEz_uno

mauro3:
Thanks for the clarification (I think you should open an issue with the test case you posted to julia-users).  Doing it the other way will need some careful designing of the state variable, but that is probably worth it in the long run.

mauro3:
Could use or take inspiration from: https://github.com/spencerlyon2/IterationManagers.jl

pwl:
I re-implemented the iterators basing on the current master, that is I took into account the hue reimplementation of solvers by @mauro3 in PR #68.  I tried my best to leave the algorithms intact but this is an initial version and there will be some bugs.  I also cut out some of the stuff, which I plan to add gradually back but first I wanted to hear out some opinions.  In particular I removed the negative direction of integration (this can be easily added as a wrapper to the current iterator protocol) and I removed most of the stuff regarding types, so at this point iterators probably don't support custom types.  On the positive side, the dense output is still supported (as a wrapper)!  There are two ways you can construct an iterator (based on e.g. `bt_feuler`), either
```
prob = ODE.solver(f,y0,t0;method = ODE.bt_feuler)
```
or
```
prob = ODE.bt_feuler(f,y0,t0)
```
(the latter overloads the `call` for `TableauRKExplicit`).  The dense output is a wrapper around the above mentioned iterators and can be constructed via
```
prob = ODE.DenseProblem(f,y0,t0; tspan = tspan, points = :specified, method = ODE.bt_feuler)
```
Sorry for the inconsistent naming convention, these are temporary placeholders anyway.  The `DenseProblem` expects the keyword `method` to be a constructor of an iterator (that is the reason for overloading the call for TableaurRKExplicit), and it passes all other keyword arguments to this constructor, so one could call it like
```
dense = ODE.DenseProblem(f,y0,t0; reltol = 1e-5, abstol=1e-3, 
   tspan = tspan, points = :specified, method = ODE.bt_feuler)
```
In principle, one could implement some other iterator (`solver`) and pass its constructor to the dense output wrapper like this.  Alternatively we can have dense output as a wrapper on an already constructed solver, then one has to construct the solver first and then pass the result to the dens output wrapper, which makes it two lines of code instead of one.

In the end we can use the new iterators as usual:
```
 for (t,y) in prob # replace prob with dense for dense output
     @show (t,y)
     if t > 3
         break
     end
 end
```

Once we fix the interface I could work on the removed features (arbitrary scalar-like types and reverse time integration).

As of yet, there is no standardized way of collecting the results but there has been a recent discussion of what type of result we expect from a solver so I decided to wait until that issue is resolved.

And of course tests don't pass at this point.

pwl:
Another update, I added the reverse time integration and some simple root finding option (via already existing Hermite interpolation in dense output).  The root finding uses a very simple bisection algorithm.  Here is an example summarizing the new features
```
import ODE

f(t,x)=[x[2],-x[1]]
y0  = [0.0, 1.0]
t0  = 0.0
tspan = -collect(1:5)
stopevent(t,y) = y[1] > 0.0

prob = ODE.solver(f,y0,t0;
                  reltol = 1e-7,
                  abstol = 1e-7,
                  tstop = -Inf,
                  method = ODE.bt_feuler,
                  tspan = tspan,
                  points = :specified,
                  stopevent = stopevent,
                  roottol = 1e-10)

for (t,y) in prob
    @show (t,y)
end
```

and the result is
```
(t,y) = (0.0,[0.0,1.0])
(t,y) = (-1.0,[-0.8414842895974096,0.5403108491723219])
(t,y) = (-2.0,[-0.909326182080129,-0.4161599958706816])
(t,y) = (-3.0,[-0.14112670311715536,-0.9900394570025749])
(t,y) = (-3.141592654651042,[3.9867758613623664e-11,-1.0000496742146774])
```

The `tspan` together with `:specified` ensure that the results will be returned at times `t=-1,-2,-3,-4,-5` (plus the initial time `t=0`, we should decide whether this is the default behavior) but before we reach time `t=-4` the root finder triggers and finds the root (the function `stopevent` returns `true`).  The tolerance for the root finder is given by `roottol`, we find that the root is `t=-pi+/-10e-9` and this is of course true because the solution is `y[1]=sin(t)`.  There are still some rough edges, for example the `tstop` and tspan are somewhat redundand.  Apart from that the support for scalar-like types is not yet there.

And there is a bonus as well: without any modifications we can use the algorithms from DASSL! (although we need a wrapper to conform to the API of DASSL)
```
using DASSL
using Iterators

dassl_wrapper(f,y0,t0;kargs...)=imap(x->(x[1],x[2]),
                                     dasslIterator((t,y,dy)->f(t,y)-dy, y0, t0; kargs...))

prob = ODE.solver(f,y0,t0;
                  (...)
                  method = dassl_wrapper,
                  (...)
```

pwl:
@tlycken @mauro3 would you mind taking a look at my implementation of the iterators?  Holiday season is coming so I might have some time to add the tests and optimize it a bit.  It could also be an opportunity to modify the output according to #82 or #80.

mauro3:
Sorry to not have looked at this yet!  I'll try over the next few days, but I suspect I may well not be able to give much input until mid-January.

tlycken:
I took a quick look, and although I won't have time to prod deeper, at least on the surface it looks quite good to me.

Do we have usage docs for this library somewhere, that should be updated?

src/dense.jl
```diff
@@ -0,0 +1,174 @@
+# iterator for the dense output, can be wrapped around any other
+# iterator supporting tspan by using the method keyword, for example
+# ODE.newDenseProblem(..., method = ODE.bt_rk23, ...)
+
```
mauro3: Shouldn't the following types have their fields typed?

src/rk.jl
```diff
@@ -0,0 +1,295 @@
...
+    reltol
+    abstol
+    minstep
+    maxstep
+end
+
+
+# overload the call method for TableauRKExplicit, it returns the
+# iterator (fixed or variable step according to the tableau)
+function call(tab::TableauRKExplicit,
```
mauro3: Is the added abstraction of using `call` justified for this?

src/tableaus.jl
```diff
@@ -0,0 +1,159 @@
+###########################################
```
mauro3: Maybe keep the RK tableaus with the RK code?

mauro3:
I had a brief look over it: looks good.  A few comments in-line, probably redundant as this is still WIP.

To test performance, maybe use https://github.com/mauro3/IVPTestSuite.jl: Threebody and HIRES should work for the explicit solvers. (I should get that package cleaned up to be more user friendly and more directly integrated...)

Random thoughts/questions:
- would multi step solvers fit into this scheme?
- would this integrate with the Sundials.jl solvers as well?  It would be cool to eventually be able to call those through ODE.jl.
- could the `start` and `done` be generic and a particular class of solvers would only need to implement `next`?
- This design assumes that dense output can be attached to the solver afterwards.  I think, for more sophisticated (higher order) dense output, the solvers and the dense-output are often more tightly coupled.  Not something to implement now but something to ponder.
- there seems to be some overlap between the types `DenseProblem` and `Problem`, maybe could be refactored?
- eventually, I would like to see in-place functions to be used in the solvers (see e.g. #72).  Would this work with this?

I'll look at it in more detail next year... thanks for the work!

mauro3:
Actually, above points about "sophisticated dense output" and multistep methods may need the same solution: a state which holds more than just the last output but either several or intermediate steps.  Pawel, as your DASSL code is multistep, maybe implementing that within this framework would lead to valuable insights on a general design (as well as more work!)?

src/rk.jl
tlycken: I'd rather have an explicitly named function than overload `call`, e.g. `solver(tab, ...)` or even `solverfor(tab, ...)`.

I also think this should be considered an internal API, and we shouldn't export it. The canonical way to construct a solver should be to call `ode`, `ode4`, `ode45` etc.

pwl:
Hi,
I made some major updates to the PR and refactored the code loosely basing on the [ideas](https://github.com/mauro3/ODE.jl/blob/m3/future-plans/doc/ideas-for-future.org) by @mauro3.  The changes include

- Support for the legacy(?) interfaces like ODE.odeXX, so most of the tests pass (but not all, there might be something wrong with the adaptive steppers).

-  Complete rewrite of the lower level structures, we now have `ExplicitODE` to hold the ode data, `Options` to keep all the options in one place and an`AbstractStepper` with the algorithms and the associated iterator structure.  Everything is held together by the `Solution{AbstractStepper}` type.  Again, the names can be changed if their meaning is not clear enough.

- Given `ode=ExplicitODE(...)` and `stepper=SomeStepper(...)` one can construct an iterator by calling 
    `sol=solve(ode, stepper; kargs...)`

- The modified Rosenbrock method ode23s is now supported

- The dense output wrapper is now a `DenseStepper` and can be used as any other stepper, to get the dense output use `dense(sol)` with `sol` as above.  The dense stepper takes its options from the `Options` component of the `Solution`

- All implemented algorithms support the in-place formulation (i.e. I switched from `dy=F(t,y)` to `F!(t,y,dy)`).  In fact this is the default way of defining an ODE now and the higher level interface converts `F` to `F!`, and does the same for the Jacobian.

Future plans

- Implement the multistep methods (I think this is the only remaining unimplemented big chunk of the existing code).

- Add tests for the lower level interface and the iterator protocol.

- Incorporate DASSL?

pwl:
I'm pretty sure there is a bug in the RK stepper (I don't get any convergence with bt_dopri5), if someone spots it please let me know.

mauro3:
Cool! I'll look at this Friday in a week (currently I'm in a pre-conference frenzy).

src/types.jl
```diff
@@ -0,0 +1,154 @@
...
+            error("Option points = $points is not supported, use :specified or :all")
+        end
+        new(initstep,tstop,reltol,abstol,minstep,maxstep,norm,maxiters,sort(tspan),points,stopevent,roottol)
+    end
+
+end
+
+
+function show{T}(io::IO, opts :: Options{T})
+    println("")
```
mauro3: just iterate over fieldnames?

src/types.jl
```diff
@@ -0,0 +1,154 @@
...
+    norm     :: Function
+    maxiters :: T
+
+    # dense output options
+    tspan    :: Vector{T}
+    points   :: Symbol
+    stopevent :: Function
+    roottol  :: T
+
+    function Options(;
```
mauro3: Maybe add the usual positional constructor, add the invariants to that and pipe the keyword constructor through that.  If this is immutable we also need an updating constructor, e.g. `Options(opts, tspan=[1,4])` to make a new options-struct with different tspan.

src/types.jl
```diff
@@ -0,0 +1,154 @@
+abstract AbstractODE
+
+
+type ExplicitODE <: AbstractODE
```
mauro3: Needs doc, in particular as there are implicit/explicit solvers and implicitly/explicitly defined ODEs.

mauro3:
Thanks @pwl, this is coming together nicely!  Big PR, big review ;-) I didn't run much code, apart from the tests.  So this is just a reading-the-code review.

In fact, can you add an example to `example/` on how to use all levels of solver API?  That would also make reviewing a lot easier as I could just run those.

My internet was patchy when reviewing this (on the train).  So instead I made the in-line comments in the code and did a PR against @pwl's fork: https://github.com/pwl/ODE.jl/pull/1/files . I hope this is ok.

We need to solve the problem with scalar types!  Otherwise this PR will be to big a regression.  I for one would be happy to make scalar types not a performance priority.  But I would hate to see them go.  What about just wrapping it in an array, why does that not work anymore?  This is what used to be in runge-kutta.jl:
```julia
function oderk_fixed(fn, y0, tspan, btab::TableauRKExplicit)
    # Non-arrays y0 treat as scalar
    fn_(t, y) = [fn(t, y[1])]
    t,y = oderk_fixed(fn_, [y0], tspan, btab)
    return t, vcat_nosplat(y)
end
```
Does that not work anymore?

What are the proposed semantics of the returned values in the iteration?  From a performance point of view, it would make sense that y is what is contained in the state itself and then would need copying before storage.  This wold allow preallocated storage to be used effectively:
```julia
yout = similar(y0, length(y0), length(tspan))
tout = similar(y0, length(tspan))
for (i,(t,y)) in enumerate(oderkf(F,y0,t0,coefs...;args...,tstop=tstop))
    tout[i] = t
    xout[:,i] = x
end
```
I think this is what is currently implemented, but we should probably make a conscious decision on that.

This needs benchmarking.  I or a GSoC student will resurrect the IVPtestsuite.jl and try to run this on there.  But I'm mostly away from the computer for about three weeks, so it might be after that.

`make_consistent_types` is not used anymore, is that on the todo or to be scrapped?

There are many types with fields which have some non-concrete type.  This makes me a bit nervous as this can introduce type instability, however, it may not matter.  Maybe comment where you use non-concrete that their use there are justified?

Including DASSL would be great, but leave that to another PR, this is huge already.

I'm not a fan of the spaces around `::` (in particular not in function signatures) and is not usual julia style.  Maybe change that?

I think (now) that `copy!(x1,x2)` is clearer than `x1[:] = x2`.  Maybe it's even faster.

I don't see why the meta-programming in interfaces.jl is needed.  I think this can be done otherwise.

Do you think it would be possible to refactor the adaptive time stepper (i.e. the next-method) such that the same can be used for ode23s and RK?  No need to do this now, just wondering.

Documentation is lacking, but I'm sure you're aware of that.  Would make reviewing easier though!

In general, this looks good, but probably needs a few more iterations until it's ready.  Which brings me to the planning: I'll be around from mid May onwards and should be able to review code within a few days of posting it.  If you got time then too, we could try and make some progress on this.

Another thing on the planning side: there is going to be a GSoC student (@obiajulu) working on ODE.jl this summer (mid May onward).  It would be great if this gets to a state that the student could base his work on this PR from sometime June onwards.  It is probably unlikely that this PR will be merged in this time-frame but if it starts to stabilize the student could base his work off your branch.

Tnx for all the work!

pwl:
The comments look as thorough as the PR itself:-).  I will take a look at them today and get back to you.

mauro3:
Cool! Well, I'm off on holidays now, so maybe no response from me for three weeks.

pwl:
The top level interface (interfaces.jl) is still a mess, but it looks slightly better now.  I really dislike the way the `initstep` is handled, just having it there requires a lot of additional code in other places.  Any ideas how to improve it?

@mauro3 The scalars were never left out, it's just that the lower-level interface doesn't support them (steppers).  Perhaps there is a way to implement it without changing the steppers but it would probably be another wrapper.  And I don't think someone willing to use the lower-level API would be disappointed to see that the scalars are not supported.  But if there is a nice solution I'm all for it.

As for the `make_consistent_types` I don't think it's necessary any more, I would just convert the Butcher tableau to `T=typeof(t0)`.

src/dense.jl
```diff
@@ -0,0 +1,193 @@
...
+# m3: I think it would be nice to factor out the dense-output and
+# root-finding into its own function.  That way it could be used also
+# independently of the dense-output iterator.  Also, it would make
+# this next function more compact.
+
+# pwl: I agree, but then the problem is that once you decouple them
+# you would lose the opprotunity to detect the roots with each step.
+function next(s::Solver{DenseStepper}, state::DenseState)
+
```
mauro3: But wouldn't you take one step with `next` which gets you to the next time, determined by the adaptive control. Then use that to create dense output and solve any equations.  So, you'd have an adaptive stepper `next` and then wrap that with a dense-output `next` which also does root finding.

src/interfaces.jl
```diff
@@ -0,0 +1,112 @@
...
+        solution = collect(dense(solver))
+    elseif issorted(reverse(tspan))
+        # Reverse the time direction if necessary.  dense() only works
+        # for positive time direction.
+
+        # TODO: still ugly but slightly less bandaid-like then the
+        # previous solution
+        solution = map(ty->(2*t0-ty[1],ty[2]),collect(dense(reverse_time(solver))))
+    else
+        warn("Unsorted output times are not supported")
```
mauro3: I think an error is better here.

mauro3:
Tnx for all the work.  I don't quite see the ugliness of the `initstep` but maybe I'm missing something. One thing to add is that `dtinit` should probably dispatch on the stepper type too.

Good to hear that the scalars are still working, sorry about the confusion.

Concerning the `make_consistent_types`: you're right, I probably over-engineered that.  There are two issues:

(1) question is whether we need to provide for a situation where `typeof(y)!=typeof(y./t)`. (This is suggested in the [API docs](https://github.com/JuliaLang/ODE.jl/blob/master/doc/api.md).)  But I can't think of any time where that would be the case. (And this is what's in `make_consistent_types`.)

(2) One reason to make the `make_consistent_types` function was to allow calls like `ode((t,y)->2t, 0., [0,1])` where `eltype(tspan)==Int`.  This should be handled graciously by the at least the high-level interface (probably should add a test for this).  I think this PR may fail on this, but I haven't checked.

Anyway, maybe `make_consistent_types` is a bit over the top and could be trimmed.

test/runtests.jl
```diff
@@ -10,19 +10,19 @@ solvers = [
            ODE.ode2_midpoint,
            ODE.ode2_heun,
            ODE.ode4,
-           ODE.ode4ms,
-           ODE.ode5ms,
+           # Ode.ode4ms,
```
obiajulu: So we are dropping the 4th and 5th order Adam-Bashforth implicit solvers, as well as the underlying `ode_ms` from `src/runge_kutta.jl` in current master, which can calculate the Adam-Bashforth coefficients for higher orders. Is this because these solvers aren't much in demand?

pwl: Sorry, I just didn't have time to implement them with iterators.  I was planning to initially so I removed the methods to add them later.  I should re-include the old implementation and leave it there.

obiajulu: No problem at all, I figured this was still in progress. I'm actually attending JuliaCon in Boston, if you are here also, it would be great to find some time and working on progressing this PR. Also, I ran some performance tests with @mauro3's IVPTestSuite.jl, and I could show you some of the results and how to run them yourself.

pwl: Great, I'm at the JuliaCon as well, how about we meet in the lobby after this lecture?

obiajulu: Sure thing.

pwl:
Now for some reason I can't make it work for `BigFloat`:
```
julia> ODE.ode4((t,x)->x,BigFloat[2,3],[1,2])
ERROR: MethodError: `__ode#7__` has no method matching __ode#7__(::Function, ::Float64, ::Float64, ::BigFloat, ::Array{Any,1}, ::Function, ::Array{BigFloat,1}, ::Array{Float64,1}, ::ODE.RKStepper{:fixed,:bt_rk4,Float64})
Closest candidates are:
  __ode#7__{T}(::Function, ::T, ::T, ::T, ::Any, ::Any, ::Any, ::AbstractArray{T,1}, ::ODE.AbstractStepper)
 in ode_conv at /home/pawel/.julia/v0.4/ODE/src/interfaces.jl:82
 in ode4 at /home/pawel/.julia/v0.4/ODE/src/interfaces.jl:70
```
I have never seen an error like that before (look at the mangled name of the `ode` function).  Something similar was already reported https://github.com/JuliaLang/julia/issues/13919 and it seems it has something to do with varargs.  If you use `BigFloat` with the third argument it hangs for a while and crashes Julia (0.4.5).

src/interfaces.jl
```diff
@@ -110,3 +122,27 @@ function reverse_time(sol::Solver)
...
+# most of the scalar operations.  In particular the element type
+# should be closed under the division.
+function make_consistent{S<:AbstractStepper}(y0, tspan::AbstractVector, stepper::Type{S})
+    t_test = 1/(tspan[end]-tspan[1])
+    Tt = typeof(t_test)
+    t_new = convert(Vector{Tt},tspan)
+
+    y_test = y0./t_test
+    Ty = typeof(y_test)
+    if typejoin(Ty,AbstractArray) == AbstractArray
```
mauro3: `Ty<:AbstractArray` or maybe better `Ty<:AbstractVector`

src/interfaces.jl
```diff
@@ -110,3 +122,27 @@ function reverse_time(sol::Solver)
...
     return solve(ode_reversed,stepper,options)
 end
+
+
+
+# The the elements of tspan should basically be scalars and support
+# most of the scalar operations.  In particular the element type
+# should be closed under the division.
+function make_consistent{S<:AbstractStepper}(y0, tspan::AbstractVector, stepper::Type{S})
+    t_test = 1/(tspan[end]-tspan[1])
```
mauro3: I think mathematically, Rationals don't make sense either, even though they are closed under division because differentiation is not defined with them.

src/interfaces.jl
```diff
@@ -110,3 +122,27 @@ function reverse_time(sol::Solver)
...
+    Tt = typeof(t_test)
+    t_new = convert(Vector{Tt},tspan)
+
+    y_test = y0./t_test
+    Ty = typeof(y_test)
+    if typejoin(Ty,AbstractArray) == AbstractArray
+        Ey = promote_type(map(typeof,y_test)...)
```
mauro3: On my computer, if `y_test` is a 10^5 element vector (say from a PDE), this takes ~23 seconds.  

I think it would make sense to check that `isleaftype(eltype(y0))` and if false issue a warning: "The `eltype(y0)=$(eltype(y0)` is not a concrete type, this degrades performance substantially."

So
```
    y_test = y0./t_test
    Ty = typeof(y_test)
    if Ty<:AbstractVector
        # if eltype(y0) is non-concrete, then eltype(y_test) will also be non-concrete
        Ey = eltype(y_test)
        if !isleaftype(Ey)
            warn("The `eltype(y0)=$Ey` is not a concrete type, this degrades performance substantially.")
        end
        y_new = copy!(similar(y0,Ey), y0)
    else
        y_new = convert(Ty,y0)
    end
```

src/runge_kutta.jl
```diff
@@ -1,480 +0,0 @@
-# Explicit Runge-Kutta solvers
```
mauro3: Why rename the file to `rk.jl`?

test/runtests.jl
mauro3: Yes, I think it's fine to leave some old-style methods and update them another day.  In fact, maybe it makes sense to keep some simple fixed-step methods as they are, to retain some easy to understand solvers.

src/interfaces.jl
pwl: My point was to avoid specifying a table here, we accept user defined types after all.  The conversion to a one element table takes place in `ode`.

src/interfaces.jl
pwl: Yeah, I wouldn't do it but if someone explicitly specifies rationals should we change that?

mauro3: Maybe not.  There is no type in Julia which signifies a representation of "continuous" real numbers.

src/runge_kutta.jl
pwl: I had no idea I did that, will fix.

src/interfaces.jl
mauro3: I'm pretty sure that `typejoin(Ty,AbstractArray) == AbstractArray` is equivalent to `Ty<:AbstractArray`.  The typejoin will produce an `AbstractArray` if `Ty` is a subtype of `AbstractArray` or `Any` otherwise.  I just think that `Ty<:AbstractArray` is more common/readable/standard.  The reason to use `AbstractVector` is that I think we should treat everything but a `AbstractVector` as scalar-like.

src/interfaces.jl
pwl: I kind of expected bad performance, the gain is that it converts the `Any[1,BigFloat(1)]` to `BigFloat[1,1]`.  With your code it would leave it as `Any`.  I'm good either way, I just wanted to point out that it's a trade off we should be conscious about.  Honestly I believe that people who are using the top level interface won't be putting huge arrays into it.  How about we test if `y_test` is of a concrete type before trying to figure out the type with `promote_type`
```
if Ty <: AbstractArray
        if isleaftype(eltype(Ty)) & (eltype(Ty) <: AbstractFloat)
            y_new = convert(Ty,y0)
        else
            Ey = promote_type(map(typeof,y_test)...)
            y_new = copy!(similar(y0,Ey),y0)
        end
    else
        y_new = convert(Ty,y0)
    end
```

pwl: And what about the bug I got?  Can you recreate it?  Do we leave it out like that?  Any idea how to fix it?

mauro3: I thought that maybe someone does indeed need to do a calculation with a `Any[]` or `FloatingPoint[]`, but in most cases it would be a mistake.  Thus the warning but not an error nor a conversion.

src/interfaces.jl
pwl: Oh, I misinterpreted that, I thought that you mean specifying the type of `y0` in the method interface.  Already fixed that in the example below.

mauro3:
I can reproduce the bug.  I suspect it's a Julia bug.  Removing the `initstep` from the signature makes it work again.

```
function ode{T}(F, y0, tspan::AbstractVector{T}, stepper::AbstractStepper;
                jacobian::Function  = (t,y)->fdjacobian(F, t, y),
                # we need these options explicitly for the dtinit
                reltol::T   = eps(T)^T(1//3)/10,
                abstol::T   = eps(T)^T(1//2)/10,
                kargs...)
    
    initstep::T = dtinit(F, y0, tspan, reltol, abstol; order=order(stepper))
...
```

mauro3:
I'm not sure this is related, are you aware of the BigFloat deepcopy issue: https://github.com/JuliaLang/julia/issues/16667?

mauro3:
Ok, so it is two separate issues: 

- if the `::T` is left off then it works but hits the BigFloat-deepcopy issue JuliaLang/julia#16667. (tested by replacing all deepcopy with copy)
- if the `::T` is used, then there is some odd methods finding error. (which persists with the deepcopy->copy replacement)

mauro3:
Ok, I found it, it's not a Julia-bug after all.  The call `dtinit(F, y0, tspan, reltol, abstol; order=order(stepper))` returns a `BigFloat` but `T==Float64`, thus it fails.  The strange mangled function name comes from internals of the keyword function machinery.

pwl:
Ok, I added a conversion to `dtinit` and it solved the issue with `y0`, but having a BigFloat `tspan` still crashes Julia 0.4.5.

pwl:
I also get a crash from
```
julia> ODE.ode4((t,y)->2y, Number[1,1.1,BigInt(1)], [0,1])
```

pwl:
To me this "feature" seems to be causing more trouble than good and `interfaces.jl` now looks like a patchwork and is difficult to follow.  How about we simply print an error if somebody tries to use `Array{Integer}` or `Integer`?  Seems much simpler to me than trying to figure out what the user had in mind?  And it's trivial to fix on the user side.  On the other hand if somebody tries to use `Array{Any}` or something similar we just print a warning and let them do whatever they like.

mauro3:
But `ODE.ode4((t,y)->2y, Any[1,1.1,BigInt(1)], [0,1])` crashes too.  How about just making the warning an error?

pwl:
So we end up supporting only leaf types and arrays of thereof?  I'm all for it.  But then we can perhaps do it with method dispatch.  Is there a type test that would filter these out on dispatch?

mauro3:
No, I don't think so.  But the check is simple and can give an informative error.

mauro3:
BTW, are you planning to work on this during tomorrow's hackathon? I could be around (remote) from about 2pm-5pm.

pwl:
That's the plan, although I believe that this is the last serious issue left out to fix and it's almost fixed now (apart from putting back the old methods).  I was also planning to work with @obiajulu to port his algorithm to the new API.

mauro3:
Cool, I'll be around.

mauro3:
I wonder a bit on how to take this forward.  As this is pretty much a full rewrite it will probably take some time (months?) to iron out all the details (bugs, API updates, etc).  I wonder whether this should first be merged into a `dev` branch on ODE.jl, where the further development could take place.  Then once it is ready merge that into `master`, this way `master` could continue to receive bug fixes.  Also having this on the `ODE.jl/dev` branch would mean, that it would only be a `Pkg.checkout("ODE", "dev")` for anyone to give this a try.  Maybe you can discuss this with @ViralBShah and others interested who are around at the JuliaCon hackathon?

pwl:
I have discussed it with @ViralBShah and he's all for merging it.  I'm also for it, this way we get more bug reports (even with your reviews we probably missed a lot of them) and possibly more contributors (people will see that this is not a stagnant package).  Also this would take this project off my hands.  Plus the top level API is not supposed to change at all (apart from some keywords).

pwl:
Can you set up an official `ODE.jl` gitter channel so we don't clutter this PR?

acroy:
+1 for a dev branch and thanks for all the hard work!

pwl:
I replaced the finite difference Jacobian with the ForwardDiff one.  It's less code and less arbitrary parameters.

ViralBShah:
I met @pwl yesterday and I loved what he told me about this PR. My reaction was to keep moving ahead with improvements, as this package is probably not too widely used.

ViralBShah:
Maybe a dev branch is the right way forward - but question - is this clearly the new way forward if we didn't have to worry about compatibility?

tkelman:
Can check METADATA for other registered packages that depend on ODE.jl, and run tests against this rewrite. If they would break with the changes here, we can make the first release that includes this a new minor or major version, then put an upper bound on the existing tags of other packages' dependencies so they will use older tags of ODE until they're updated.

src/rosenbrock.jl
```diff
@@ -0,0 +1,89 @@
+#ODEROSENBROCK Solve stiff differential equations, Rosenbrock method
```
mauro3: These are the old solvers, right?  If so, I'd re-add them to `src/ODE.jl` where it was.

src/runge-kutta.jl
```diff
@@ -0,0 +1,315 @@
+# This file contains the implementation of explicit Runkge-Kutta
```
mauro3: Nice try ;-) You need to rename it with `_` : `runge_kutta.jl`

src/rosenbrock.jl
pwl: Why not keep them in separate files?  That's what we do with runge-kutta and modified Rosenbrock.

mauro3: Because nothing changed with them. Then leaving them in the place they were, the diff will show that indeed nothing changed.  This is a huge PR already, so anything stable is appreciated as that is less to review.  Same goes for the Adams solvers.  Once they get updated (which should be a separate PR) then I'm all for it to have them in a separate file.

mauro3:
@ViralBShah: The reason to make this a `dev` branch is not breaking high-level API (which it doesn't) but just that it is still very experimental but could profit from user testing.  My understanding is that ODE's master branch is reasonably stable, whereas this isn't.  I think the call on whether this is the way forward could be made later, latest when it is ready to be merged into master.  Also, as @pwl said, it could maybe improve ODE.jl's reputation and show that progress is indeed being made (see this [julia-users thread](https://groups.google.com/d/msg/julia-users/QMjL1EaKnpQ/aft2q3N5BAAJ)).  

If this is indeed turned into a `dev` branch, then it might make sense to give @pwl and me write access to facilitate further development.  Otherwise, it is probably better to keep it on a fork.

@tkelman: testing this PR against packages would be great! Maybe check back with @pwl on when's a good time.  Also, this PR doesn't seem to trigger Travis, last did half a year ago, any idea why? (On my machine, it currently passes on 0.4 but fails on 0.5).

tkelman:
Most likely due to the conflicts. CI generally only runs on PR's that are merge-able

mauro3:
I see, thanks!

src/runge-kutta.jl
```diff
@@ -0,0 +1,315 @@
...
+"""
+
+A general Runge-Kutta stepper (it cen represent either, a fixed step
+or an adaptive step algorithm).
+
+"""
+immutable RKStepper{Kind,tab_name,T<:Number} <: AbstractStepper
+    tableau::Tableau
+    function RKStepper()
+        tab = eval(tab_name)
```
mauro3: I don't like this `eval`.  It needn't be changed urgently but maybe add a `TODO: remove eval`.

pwl: If you have an idea how to avoid it I'm all ears.  I thought about it and I couldn't come up with a better solution.

mauro3:
After some Gitter chat with @pwl, we agreed that it's fine to keep this on @pwl's fork for a bit longer.  We'll ping people again if a `dev` branch is called for.  Sorry for the noise, maybe I got a bit excited due to JuliaCon (even though I'm 1000s of miles away!).

src/runge-kutta.jl
tlycken: I haven't been active here for a long time, so I might have missed something here, but how about having a (const) dictionary holding the tableaus, instead of defining them as constants? That way you could do `tab = tableaus[tab_name]` instead.

pwl: That sounds good but I would rather not touch the Butcher tables until we are done with more important issues.

pwl: BTW, why is `S` a parameter of the Tableau type?  Can't it be a type field instead?

mauro3: Combining the suggestions is good: leave the tableaus alone (at least for now) but collect them in a Dict. 

I think the reason I made the `Name` a type parameter was to allow someone to write a specific stepper for a specific RK method, which has not happened yet and may never will.  Nonetheless, I think it's ok and Julian to provide that possibility.  I'm not sure why `S` became a parameter, probably should be removed at some point.

pwl: I already removed the `S` and `Name` (although I haven't commit the changes yet).  This way you can specify all the type parameters (only one remaining) in `Tableau` in `RKStepper`.  I don't think that the `Name` in `Tableau` will ever be necessary now that we have the stepper as a type.  The `RKStepper` looks like this

```
immutable RKStepper{Kind,Name,T} <: AbstractStepper{T}
    tableau::Tableau{T}
    function RKStepper()
        tab = convert(T,eval(Name))
        if Kind == :fixed && isadaptive(tab)
            error("Cannot construct a fixed step method from an adaptive step tableau")
        elseif Kind == :adaptive && !isadaptive(tab)
            error("Cannot construct an adaptive step method from an fixed step tableau")
        end
        new(tab)
    end
end
```

and you can find the `Tableau` is here https://gist.github.com/pwl/50c705538d6c86087180f99fac1bfe68 (sorry, no `Dict` yet, I can add it later).

I can still revert the changes although to me the code looks less cluttered now.  I'm not commiting anything yet because I am still looking for a fix to the type instability.

src/ode23s.jl
```diff
@@ -56,15 +56,15 @@ function show(io::IO, state :: RosenbrockState)
...
+function start{O<:ExplicitODE,S<:ModifiedRosenbrockStepper}(s::Solver{O,S})
     t  = s.ode.t0
     dt = s.options.initstep
     y  = s.ode.y0
     dy = zero(y)
 
     J  = Array(eltype(y),length(y),length(y))
 
-    step  = Step(t,deepcopy(y),deepcopy(dy))
+    step  = Step(t,copy(y),copy(dy))
```
pwl: For some reason `deepcopy` is not type stable:
```
julia> function f()
           x=[1.]
           y=copy(x)
           z=deepcopy(x)
       end
f (generic function with 1 method)

julia> @code_warntype f()
Variables:
  x::Array{Float64,1}
  y::Array{Float64,1}
  z::Any
  ######dims#1369#7034#7035::Tuple{Int64}
[...]
```

mauro3: It is type-stable on Julia 0.5.  Actually, I think it makes sense to tune this PR to 0.5 as it will be out by the time this is finished.  In particular the new functions will have a major impact on how ODE.jl performs. (In a pinch, we could even drop 0.4 support.)

pwl: I would rather leave it as is and change it later, maybe once 0.5 is out  What do you think about the new version of Tableau?

mauro3: 0.5 is pretty stable now.  Only a few more changes will go in before a release candidate is out.  It just would be a shame to spend a lot of time performance tuning, only to do it again in a few weeks.

obiajulu:
As some of you may know, Mauro and I have been working on improving the [IVPTestSuite](github.com/mauro3/IVPTestSuite.jl) package for benchmarking ODE solvers. I threw together a terminal friendly version, which can be used to run some basic benchmark test from the Julia terminal (currently, more extensive testing and plot generations can be run as scripts, requiring light modification of the innards of the package to narrow in on specific solvers and/or test cases). I thought a terminal version would make testing this, and other PRs in the future, easier. Here is how to run it: . 1) First, checkout the terminalmode branch of IVPTestSuite with `Pkg.checkout("IVPTestSuite","terminalmode")`. 2) Next, start `using IVPTestSuite.QuickSuites` as well as `using ODE`. Then you can simply run something like `runtestsuite(ODEsolverfns=[ODE.ode1,ODE.ode23s], sundialsolverfns = [], DASSLsolverfns = [])` (or `runtestsuite(ODEsolverfns=[ODE.ode1,ODE.ode23s])` since the last two default to an empty array anyway). All in all, on could run ```Pkg.checkout("IVPTestSuite","terminalmode")
using ODE
using IVPTestSuite.QuickSuites
runtestsuite(ODEsolverfns=[ODE.ode1,ODE.ode23s])```
And, if one has @pwl's PR as a local branch of their ODE repo, then you can quickly test between the master branch of ODE and @pwl's PR by 
```Pkg.checkout("IVPTestSuite","terminalmode")
using ODE
using IVPTestSuite.QuickSuites
runtestsuite(ODEsolverfns=[ODE.ode1,ODE.ode23s])
Pkg.checkout("ODE","[name of pwl's PR branch]")
runtestsuite(ODEsolverfns=[ODE.ode1,ODE.ode23s])```
Hope this can help with debugging specific solvers.

**Comment by mauro3** [comment]
Thanks @obiajulu.  I haven't checked it but I think a restart is needed in your last example:
```julia
using ODE
using IVPTestSuite.QuickSuites
runtestsuite(ODEsolverfns=[ODE.ode1,ODE.ode23s])
### restart Julia
Pkg.checkout("ODE","[name of pwl's PR branch]")
using ODE
runtestsuite(ODEsolverfns=[ODE.ode1,ODE.ode23s])
```
(BTW if you add a `julia` to the first set of backticks: ```` ```julia ````, then you get syntax highlighting)

src/tableaus.jl
```diff
@@ -78,28 +76,13 @@ function TableauRKExplicit(name::AbstractString, order::(@compat(Tuple{Vararg{In
...
-            newflds = tuple(newflds..., conv_field(Tnew, fld))
-        else
-            newflds = tuple(newflds..., fld)
-        end
-    end
-    TableauRKExplicit{Tnew}(tab.name, tab.order, newflds...) # TODO: could this be done more generically in a type-stable way?
-end
-
+Base.convert{Tnew<:Real,T}(::Type{Tnew}, tab::TableauRKExplicit{T}) =
+    TableauRKExplicit{Tnew}(tab.name, tab.order,
```
tkelman: shouldn't convert return the type it's converting to? this is a strange method, asking for a Real and getting back a tableau?

pwl: Good point, fixed.

src/ode23s.jl
```diff
@@ -0,0 +1,153 @@
...
+
+            return ((step.t,step.y), state)
+        else
+            # continue with the decreased time step
+            dt = dtnew
+        end
+
+    end
+
+    return tout, yout
```
mauro3: This `return` is not needed.

dpsanders: Many people these days in fact prefer to have an explicit `return`. It arguably makes it more immediately obvious what is happening.

mauro3: Sorry, I meant to say, that it is misleading. The function should only ever return `return ((step.t,step.y), state)`, but execution never gets there, so it doesn't matter.  

I agree about the explicit return statement.

ivarne: The term I often see used is `dead code` or `unreachable code`. Sometimes type inference isn't able to determine that controlflow never reaches the end of a function, so you need an extra return statement so that you get better type information in the calling function. (An exception would have the same effect.) Anyway you should include a comment to make the purpouse of the return (or exception) obvious to the reader.

mauro3: Thanks! In that case an exception would be best.

src/base.jl
```diff
@@ -99,6 +99,8 @@ Subtypes include: `AbstractIntegrator`s but also `DenseOutput`
 """
 abstract AbstractSolver
 
+legnth(s::AbstractSolver) = error("`length` is not defined for $(typeof(s)).")
```
helgee: Typo ;-) `legnth -> length`

ChrisRackauckas:
Hey, talked with @pwl last night in depth about this change. I think that, when it comes time, we should

1. Tag the master before merging this.
2. Fork the master to another package (LegacyODE.jl?)
3. Pull this in.
4. Give it a some time to be thoroughly tested.

The reason for 2. is because the current source code has good educational purposes and is very approachable. It's pretty much what people would think the standard implementation for ODE solvers would be. On the other hand, while there is a legacy API to make it act the same (though the idea is to deprecate this: JuliaODE/ODE.jl#25), the iterator format definitely has a bit of a barrier to entry. And DifferentialEquations.jl has a lot of extra stuff in its algorithms which are related to getting good performance which also makes it not so good in terms of readability (I definitely put performance > elegance whenever there's a tradeoff). So I think going forward that it would be good to keep this (and with the next few updates) master around as a different package which touts the legacy `odexx` interface and is easily hackable.

Because of this, at first I suggested that maybe this spawn off to another package, but @pwl convinced me that it's a good idea for ODE to take up this change since otherwise ODE will probably lose most of its development activity (i.e. @pwl, @mauro3, and everyone else interested in the iterator API) and its probably best to choose the path where ODE gets the most development. 

Thoughts?

ChrisRackauckas:
On a different note, I think I should mention that this branch does have a bunch of extra tests. Since I choose to wrap the iterator version of ODE from here into DifferentialEquations.jl, I have been following this branch pretty closely and have a good amount of my own tests for it. I also have some benchmarks, but since I never wrapped the current ODE, no regression tests (though if we are going to go forward with the LegacyODE idea, I can wrap the master as well and then we'll have a bunch of regression benchmarks. Otherwise it's not worth doing if that interface is going to disappear). 

Remind me to share these when it comes closer to merge time.

REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.5
```
tkelman: this won't work until final is released

ChrisRackauckas:
Hey,
  I was wondering if we should talk about this becoming a separate package. There's a few reasons for this:

1. This a huge breaking change, essentially a new package. It seems that almost all (if not all) of the code has been modified in some way, and the API is different. No previous code or examples (which there are quite a few examples if you Google around) will work with this library after the change. Although for this kind of library there aren't that many dependents, I think that judging the usage by dependents understates its usage as seen by Googling for examples. Besides, it already has its own specific issues, PRs, etc.
2. This is a large enough change that it basically invalidates the other PRs for this repository. Other people have some interesting contributions which could be added.
3. @tkelman has mentioned a few times that he would like to move away from the name ODE.jl. Going out of the way to ensure this other code essentially becomes ODE.jl is then clearly the wrong direction.
4. [pwl already mentioned that he will not be maintaining this in the future](https://gitter.im/JuliaDiffEq/Lobby/archives/2016/10/28). This is a really large change and the API is much more involved, and I'm not as comfortable with replacing all of the code in a public package with one for which the main author is no longer present.
5. This does add dependencies. Not many, but a few. This is something to think about if we want a low-dependency library, of which ODE.jl is the only one.
6. There's no problem with making a new package. We can transfer JuliaODE/ODE.jl into JuliaDiffEq with a new name (ODEIterators.jl?) and register it tonight with no problems, and it could be updated as needed. The only issue with why it needs to be so perfect for the PR is that it's re-writing the entire codebase of an already used public package, so it must cover all loose ends. Making it a new package is also less misleading to users.

The end result of making it a separate package is that, not only would it be easier on @pwl, it would also be less confusing to current users and offer a a completely different design path with different advantages and disadvantages (the upvotes for spinning the current off to LegacyODE.jl shows that there people do see upsides to the current ODE.jl architecture and want to keep it intact). 

I'd like a second opinion. @ViralBShah?

pwl:
The main reasons for replacing ODE.jl is that the new version would be easier to maintain, with less code duplication and more documentation.  But I'm not opposed to the idea of making it a separate package as you suggest (see point 6. below).

1. We designed the new version to be backward compatible, there might be some failing edge cases but replacing the current ODE with our ODE shouldn't cause any code to break (at least that's what we were aiming for).
2. I agree on this front.  Just a comment: some of these PRs are already implemented in our version, so there is no point to hold on to them too much.
3. I had no idea about that.  Does it mean that ODE.jl will be renamed?  Where did this discussion take place?
4. Even if I won't have too much time to work on it it doesn't matter that other people wouldn't be able to contribute, especially @mauro3 .
5. The only dependency that we add is `ForwardDiff`, which any package implementing implicit methods should use in place of finite difference approximations.  And it's the first time I hear about the need for a low-dependency library for solving ODEs.
6. That is a completely valid point and an alternative solution to a somewhat stagnant situation we have now.  Here is my take on it: instead of replacing ODE.jl we leave it as is but deprecate it, which would encourage people to use `ODEIterators.jl`.  This way we could keep the old package with all the issues, feature requests and PRs and let the users gradually move to a different package.  Because the API is backward compatible (to a degree) it should be as easy as changing `using ODE` with `using ODEIterators`.  What do you think about that?  I'm not too sure about deprecation but it might have a positive catalyzing effect and it would speed up the adoption of `ODEIterators`, it's certainly worth discussing.

mauro3:
I think, ideally, there should be one main, fully Julia initial-value-problem package (the equivalent of Optim.jl in the JuliaOpt universe).  The obvious choice is ODE.jl.  However, @ChrisRackauckas is right that this PR has become too big for an easy merge. I would favor a split into LegacyODE.jl and ODE.jl as I think ODE.jl is a good name (if somewhat misleading) and that would make it clear on which package to focus development.  IVP.jl would be a more correct alternative but the abbreviation is not as well known.  What were @tkelman's objections to the name ODE.jl?

Concerning the points:

1) The idea is that the old syntax would still be working.  This has been relaxed a little but 99% of old code should be running with this PR.  I just checked the three examples and they run (except that `ode45` is not exported thus needs `ODE.ode45`).

2) True.  But if the old API is still valid then the old solver should just be copied and the ported at a later stage.

4) Yep, this is the main issue.  It would be necessary to have community support for this change to be worth-while.  At the moment this basically means you, @ChrisRackauckas.  I think this maybe better discussed in a roadmap issue, I'll put one together in https://github.com/JuliaDiffEq/Roadmap.

5) This adds ForwardDiff.jl, which is ok for me.  But I agree dependencies should be kept to a minimum and Julia-only packages.

6) Moving this to JuliaDiffEq would be good.  We could just rename it to something that makes it clear that it is in development and postpone the decision about package names to when it becomes operational.

mauro3:
Also, just having had a look through JuliaDiffEq: is there a point in having both https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl and ODE.jl (or a new incarnation thereof)?

Also @tkelman's reservation about ODE.jl is that acronyms should be avoided: https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/59#issuecomment-254069632

ChrisRackauckas:
@tkelman spells it out more here: https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/59#issuecomment-254072605

>abbreviations are usually okay, but ODE and SDE are bad names that probably would be spelled out more if registered today. people not in the field will want to know what the package is for and possibly use it as well. longer names are clearer.

If that's the way all of Julia's naming is going, we should take that into consideration.

>ideally, there should be one main, fully Julia initial-value-problem package (the equivalent of Optim.jl in the JuliaOpt universe)

There's a lot of signs pointing that this is the wrong direction. Indeed, even Optim.jl is splitting out functionality things like LsqFit.jl to get everything ease to document and add choices which reduce dependencies. The more recent approach is the individual components + metapackages approach exemplified by Plots.jl and JuliaML's Learn.jl. This is the direction that DifferentialEquations.jl has gone, following their lead. The reason is because it allows for an "ease of use" package which is fully documented, but then individual components for developers to pull as needed, and allows for dev docs/issues to be more targeted and thus have a lower barrier to entry.

>I don't think that in the long run there is a point in having both OrdinaryDiffEq.jl+AlgebraicDiffEq.jl and ODE.jl

Depends. Sure, if you're just looking for one interface that does it all (and wraps everything else), has a bunch of features, and also offers the best possible performance in order to be easily usable for problems like tough PDEs or as a backend to DDE solvers, OrdinaryDiffEq.jl already does that. It also has a unified interface with SDEs, DAEs, PDEs, and in the near future, DDEs. If that's all that one's looking for (which is likely most users), then sure there's no point.

But I do see that there are other things which can be offered. For example, the OrdinaryDiffEq code has a lot of extra stuff in it because it provides adaptive stepsize (PI controlled) stabilization, specific dense outputs for different algorithms to match orders with minimal calculations, autodifferentiation using ForwardDiff and root solving using NLsolve. It also allows for other things like using the symbolically calculated inverse Jacobian from ParameterizedFunctions.jl to speed up stiff solvers. This adds complexity and dependencies. On the otherhand, ODE.jl is essentially dependency-free and very simple code. This PR has an interesting new iterator interface at the expense of much more complex code (at least to me, I find this much harder to follow). Of course, I am wrapping it all in OrdinaryDiffEq anyways so it is all accessible via one package and one function (currently only this PR is wrapped, but if they end up as separate packages with different implementations I'll extend the wrapper to both since that would take less than an hour), but it would offer the choice of using a separate component if necessary.

While I don't value the iterator interface myself, I think it's an interesting idea. But in the end does not get anything new (at least for diffeq solvers). It introduces a lot of code complexity in order to allow the interface, but I don't think that interface is very usable itself. While the premise is that you can "inject code", there's not enough available and so callback functions actually end up having more functionality for declaring things like events. If the algorithms have a lot of specific codes (specific dense output functions, differing amounts of caches, etc.), then using the iterator interface to something non-trivial like dynamically resizing the ODE is very difficult to do and is algorithm-specific (you change the integrator and then you have to change the code in the loop part?), so it's better to write a DSL and some macros anyways. So in the end I think I've shown that the [callback + event DSL approach](http://juliadiffeq.github.io/DifferentialEquations.jl/latest/man/callback_functions.html) is more performant and easier to use. So I don't plan on adding an iterator interface to OrdinaryDiffEq, so it's perfectly fine to have another setup which has it available. I will still allow calling whatever other methods are out there, even if it doesn't use the full iterator format (and, if it's useful, can probably put a type of callback function in the iterator part, so the same functionality is available).

>So, ideally we'd distill the best out of all of those packages and put it into one (or two)

I did pull everything I could find which was salvageable from ODE.jl already. For example, the Rosenbrock implementations served as good inspiration. However, other things like `ode78` I don't see much use for (it's a Fehlberg method, and there are much more efficient methods these days). The benchmarks show that I pulled the best out because, well, if they didn't I'd keep working on them haha (they do show that OrdinaryDiffEq's native algorithms do have about a millisecond extra of startup time over the ODE.jl algorithms due to dynamic dispatch on the `alg` symbol. So if a problem is solved in less than milliseconds, you can measure ODE.jl as faster (sometimes, since there's a lot of jitter here), but other than that the ODE.jl algorithms are slower (this is measured against this PR)). That's why I personally am going the way of just working on OrdinaryDiffEq, modularizing, and adding dev docs to help pull other contributors in. I think it's much easier to get OrdinaryDiffEq "more public" than it is to get ODE/ODEIterators to have the functionality of OrdinaryDiffEq.

However, given that its structure can easily call outside packages, there's no limitation that others have to work directly inside of it as well. It's allows for a "work as you wish" strategy. I don't really see a problem with having some algorithms in a package which is separate from the main solver, as long as it's in JuliaDiffEq so that way it can be maintained (fix depwarns, change which versions of testing, etc., even if I'm not working on it directly, I like to keep it all up-to-date and working)

PR #53, opened 2015-01-29
Need smaller hmin in ode23s to run ROBER test case

mauro3:
To run the ROBER testcase from
http://www.unige.ch/~hairer/testset/testset.html
hmin <= 1e-17

See new testcase in test/runtest.jl


acroy:
Thanks for adding the test! I don't see a problem with this change. It might be good to finally make `hmin` a keyword with some more reasonable default value (I think Hairer uses `10e-15` in RODAS)...

mauro3:
Yes, a semi-unified set of keywords would be good.  Thanks.

PR #55, opened 2015-02-18
Fix type stability for tout in oderkf and use push! instead of concatenation. 

andreasnoack:
Use vcat instead of old range to vector convertion syntax.

This pr does break 0.3 compatibility and therefore supersedes #54.


jiahao:
If the syntax will break 0.3 compatibility then `REQUIRE` should be updated with `julia 0.4-`

tkelman:
Can this be handled by Compat yet? This isn't a very good reason to make this package 0.4-only.

andreasnoack:
No. I forgot an important negation. The reason I made this new pull request was to avoid the breakage for 0.3.

Issue #30, opened 2014-03-12
Inspiration from PETSc ODE/DAE solvers for API and algorithms


mauro3:
This is to propose to support the API PETSc uses for its ODE/DAE solvers.
I think the PETSc framework is the nicest around and the most modern (afaik).
(Other issues related are #20, #18, https://github.com/JuliaLang/julia/issues/75)

The PETSc ODE/DAE solvers have been implemented over the last few years with modern developments in mind (they are citing papers from 2003, 2009), which is more than can be said for most other suites of ODE solvers which carry decades old ballast around. It supports, explicit, implicit and IMEX methods for ODEs and DAEs all within the same API. The API is based around formulating the problem like

```
F(t, u, u') = G(t, u)
```

For problems using implicit methods this is used as `F(t, u, u') =0` for
explicit ones to `u'=G(t, u)` (i.e. `F=u'`), IMEX mixes the two at once.

Best have a look at the PETSc manual which gives a nice concise intro (Chapter 6): 
http://www.mcs.anl.gov/petsc/petsc-current/docs/manual.pdf

PETSc has a liberal licence so code could probably be translated
from PETSc into Julia. The problem with that approach is that
one needs to understand the whole PETSc framework, otherwise
the code looks quite obfuscated.

Further, some cool solvers PETSc has:
- General linear methods:
  http://www.mcs.anl.gov/petsc/petsc-3.4/docs/manualpages/TS/TSGL.html#TSGL
  These methods contain Runge-Kutta and multistep
  schemes (e.g. BDF) as special cases. As far as I can tell with
  my limited ODE/DAE solver knowledge, these are the best
  currently available.
- IMEX: implicit explicit methods
  http://www.mcs.anl.gov/petsc/petsc-3.4/docs/manualpages/TS/TSARKIMEX.html
  "These methods are intended for problems with well-separated
  time scales, especially when a slow scale is strongly nonlinear
  such that it is expensive to solve with a fully implicit 
  method."

Disclaimer: I have not actually used the PETSc solvers, just read
the docs and wished I could use them in Matlab...

11718pauld:
https://github.com/hplgit/odespy

PR #57, opened 2015-03-23
Add passing of parameters to `ode23`

dpsanders:
This adds to `ode23` the  possibility to pass parameters through to the function to integrate, via parameter splatting. This is useful functionality; see, for example, [this post on Julia-users](
https://groups.google.com/forum/#!searchin/julia-users/Error$20array$20could$20not$20be$20broadcast$20to$20a$20common$20size/julia-users/4Yj2CnvPB_s/5gsUhsRyaiEJ).

I also updated (read: corrected) the documentation to use the Julia, not MATLAB, convention, and to not give `InexactError`s.

The last commit updates the documentation to use the `@doc` style.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,4 @@
 julia 0.2
 Polynomials
+Docile
+
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -19,38 +19,42 @@ export ode4s, ode4ms, ode4
 #    ode4ms, ode_ms
 
 
-#ODE23  Solve non-stiff differential equations.
+# ode23: Solve non-stiff differential equations.
 #
-#   ODE23(F,TSPAN,Y0) with TSPAN = [T0 TFINAL] integrates the system
-#   of differential equations dy/dt = f(t,y) from t = T0 to t = TFINAL.
-#   The initial condition is y(T0) = Y0.
+#   ode23(f, y0, tspan) with tspan = [t0, t_final] integrates the system
+#   of differential equations dy/dt = f(t,y) from t = t0 to t = t_final.
+#   Here, y may be a scalar or a vector.
+#   The initial condition is y(t0) = y0.
 #
 #   The first argument, F, is a function handle or an anonymous function
 #   that defines f(t,y).  This function must have two input arguments,
-#   t and y, and must return a column vector of the derivatives, dy/dt.
+#   t and y, and must return a vector of the derivatives, dy/dt.
 #
-#   With two output arguments, [T,Y] = ODE23(...) returns a column
-#   vector T and an array Y where Y(:,k) is the solution at T(k).
+#   ode23 returns the pair (tout, yout),
+#   where tout is the vector of times and yout an array of solutions:
+#   yout[k,:] is the solution at time tout(k)
 #
-#   More than four input arguments, ODE23(F,TSPAN,Y0,RTOL,P1,P2,...),
-#   are passed on to F, F(T,Y,P1,P2,...).
+#   Parameters may be passed through to the function F by adding additional
+#   arguments:
+#   ode23(f, y0, tspan, p1, p2, ...)
+#   calls f(t, y, p1, p2, ...).
 #
-#   ODE23 uses the Runge-Kutta (2,3) method of Bogacki and Shampine (BS23).
+#   Keyword arguments reltol and abstol specify the relative and absolute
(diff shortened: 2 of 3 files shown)


acroy:
Thanks for the PR. Unfortunately, things are a bit complicated ... Firstly, `ode23` will eventually be replaced by `ode23_bs` and we just kept it to keep track of regressions etc (see #16). Secondly, although we never discussed passing parameters to the rhs function (AFAIK, but maybe #30 contains something on this), the suggested solution is probably to use [anonymous functions](http://julia.readthedocs.org/en/release-0.3/manual/functions/?anonymous-functions#anonymous-functions), i.e. in your new test
```
for α in 1.0:1.0:10.
    t,y=solver( (t,y)->-α*y, 1., [0:.001:1;])
end
```
should work already now for all solvers (maybe [FastAnonymous.jl](https://github.com/timholy/FastAnonymous.jl) might be worth a look in this context). But maybe I missed what you were trying to achieve ...
Thirdly, if we would make this change we would have to do it for all solvers to keep the API more or less consistent.

dpsanders:
Thanks for the comments (and apologies for the delay in replying).

I agree that anonymous functions provide a solution to the problem, but it seems strange if one already has a function, say `SIR`, to have to wrap it in another function and write
`tout, yout = ode23( (t, x) -> SIR(t, x, a, b), x0, tspan)`
instead of just
`tout, yout = ode23(SIR, x0, tspan, a, b)`

As you imply, this will also presumably involve a performance hit, though `FastAnonymous.jl` could indeed be a solution. I would like to do a performance comparison -- are there any particularly nasty ODEs I could test this on that you usually use? 

For what it's worth, both `MATLAB` and `scipy.integrate` allow passing through parameters to the function (if I am not mistaken), and it seems a strange omission.

I am happy to go through and add parameter splatting to the other functions if this is the consensus -- I just thought that for my first PR I would be conservative!

Any thoughts on the documentation changes? Some of those are necessary (since the current documentation for `ode23` seems to just be a copy of the MATLAB version and is actually wrong and misleading, as witnessed in the post in Julia-users referenced in the PR).

mauro3:
Maybe you can use https://github.com/mauro3/IVPTestSuite.jl for performance testing.

dpsanders:
Great, thanks!

pwl:
For what my opinion is worth, I would prefer to keep using the anonymous functions to pass parameters.  This way the implementation of `ODE.jl` stays free of clutter like `params...` and is more readable.  It would be especially messy to add `params` to some more complex algorithms.  That said, I have never investigated the performance hit.

acroy:
Personally, I don't see a problem with anonymous functions (performance aside, but this will certainly improve). AFAIK MATLAB dropped passing parameters the way you suggest (at least there is noting in the [current documentation](http://www.mathworks.com/help/techdoc/ref/ode23.html)) and Python might do it for performance reasons, but someone else has to comment on the reasons. 

I would support a solution that is more general, i.e. in the direction of having a "problem" type like we discussed for instance in #30. Having this could obviously include support for passing extra parameters.

> Any thoughts on the documentation changes?

I like the changes and in particular the move to Docile. As I said above, `ode23` will likely be removed though.

mauro3:
Regarding the issue.  I prefer an anonymous function solution.  This is also what matlab does.  I'm not sure this pull request will have much impact on performance as passing functions to functions is already bad. Hopefully, this will be fixed in Julia soon!  Until then FastAnonymous.jl probably is fastest, faster than passing in the function.

tlycken:
Yeah, I think anonymous functions is much cleaner too (I actually think the `ODEProblem`-type solution is kind-of clunky, but I guess that's a matter of taste.)

Performance-wise, this is not something we should worry about in ODE.jl - Julia as a whole will be crippled compared to its ambitions until this is resolved, so I would be very surprised if this isn't fixed way before 1.0.

stevengj:
The problem with anonymous functions is that you can only do one thing with them (call them).  If the user defines an `ODEProblem` type, this is more flexible because you can define multiple methods on that type to provide function values, Jacobians, etcetera.   That is why it is good to have type-based low-level interface.  (You can still have a high-level interface based on passing simple functions for the simple cases.)

acroy:
@dpsanders : Now we have removed the old implementation of `ode23`. If you are still up for it, a PR for changing the documentation to Docile would be great.

PR #43, opened 2014-06-25
[WIP] adding output at specified points

berceanu:
I just added the `initstep`, `minstep` and `maxstep` keywords specified in the API.
Output at specified time-points is still wip, as for a `tspan` of `[0.:10.:50.]` I currently get output at

```
tout = [0.0 6.0e-7 2.351e-5 0.0005 0.009 10.013 19.981 30.016 39.983 49.997 50.0]
```

PR #59, opened 2015-03-24
Add keywords minstep, maxstep and initstep to ode23s and oderkf

acroy:
As the title says this PR adds the keywords `minstep`, `maxstep` and `initstep` to `ode23s` and `oderkf` as suggested in the API docs. The initial step estimator is actually due to @berceanu (I hope you are Ok with me taking it from #43; otherwise just let me know). The other commits are just minor cosmetic changes.

### DIFF
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -18,6 +18,32 @@ export ode4s, ode4ms, ode4
 #    oderosenbrock, ode4s, ode4s_kr, ode4s_s,
 #    ode4ms, ode_ms
 
+# estimator for initial step based on book
+# "Solving Ordinary Differential Equations I" by Hairer et al., p.169
+function hinit(F, x0, t0, p, reltol, abstol)
+    tau = max(reltol*norm(x0, Inf), abstol)
+    d0 = norm(x0, Inf)/tau
+    f0 = F(t0, x0)
+    d1 = norm(f0, Inf)/tau
+    if d0 < 1e-5 || d1 < 1e-5
+        h0 = 1e-6
+	else
+        h0 = 1e-2d0/d1
+	end
+    # perform Euler step
+    x1 = x0 + h0*f0
+    f1 = F(t0 + h0, x1)
+    # estimate second derivative
+    d2 = norm(f1 - f0, Inf)/(tau*h0)
+    if max(d1, d2) <= 1e-15
+        h1 = max(1e-6, 1e-3*h0)
+	else
+        pow = -(2. + log10(max(d1, d2)))/(p + 1.)
+        h1 = 10.^pow
+	end
+	h = min(100.0*h0, h1)
+end
+
 
 #ODE23  Solve non-stiff differential equations.
 #
@@ -186,7 +212,11 @@ end # ode23
 # CompereM@asme.org
 # created : 06 October 1999
 # modified: 17 January 2001
-function oderkf(F, x0, tspan, p, a, bs, bp; reltol = 1.0e-5, abstol = 1.0e-8)
+function oderkf(F, x0, tspan, p, a, bs, bp; reltol = 1.0e-5, abstol = 1.0e-8,
+                                            norm=Base.norm,
+                                            minstep=abs(tspan[end] - tspan[1])/1e9,
+                                            maxstep=abs(tspan[end] - tspan[1])/2.5,
+                                            initstep=0.)
     # see p.91 in the Ascher & Petzold reference for more infomation.
     pow = 1/p   # use the higher order to estimate the next step size
 
@@ -196,9 +226,14 @@ function oderkf(F, x0, tspan, p, a, bs, bp; reltol = 1.0e-5, abstol = 1.0e-8)
     t = tspan[1]
     tfinal = tspan[end]
     tdir = sign(tfinal - t)
-    hmax = abs(tfinal - t)/2.5
-    hmin = abs(tfinal - t)/1e9
-    h = tdir*abs(tfinal - t)/100  # initial guess at a step size
+    hmax = maxstep
+    hmin = minstep
+    h = initstep
+    if h == 0.
+      # initial guess at a step size
+      h = hinit(F, x0, t, p, reltol, abstol)
+    end
+    h = tdir*min(h, hmax)
     x = x0
     tout = [t]            # first output time
     xout = Array(typeof(x0), 1)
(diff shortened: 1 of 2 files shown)


acroy:
Anyone?

src/ODE.jl
```diff
@@ -18,6 +18,41 @@ export ode4s, ode4ms, ode4
...
+###############################################################################
+
+# estimator for initial step based on book
+# "Solving Ordinary Differential Equations I" by Hairer et al., p.169
+function hinit(F, x0, t0, p, reltol, abstol)
+    tau = max(reltol*norm(x0, Inf), abstol)
+    d0 = norm(x0, Inf)/tau
+    f0 = F(t0, x0)
+    d1 = norm(f0, Inf)/tau
+    if d0 < 1e-5 || d1 < 1e-5
```
tlycken: The indentation in this `if` statement and the next makes it more difficult to understand the control flow here, IMO.

src/ODE.jl
```diff
@@ -18,6 +18,41 @@ export ode4s, ode4ms, ode4
...
+    f1 = F(t0 + h0, x1)
+    # estimate second derivative
+    d2 = norm(f1 - f0, Inf)/(tau*h0)
+    if max(d1, d2) <= 1e-15
+        h1 = max(1e-6, 1e-3*h0)
+	else
+        pow = -(2. + log10(max(d1, d2)))/(p + 1.)
+        h1 = 10.^pow
+	end
+	h = min(100.0*h0, h1)
```
tlycken: This should also not be indented, to make it clerarer that `h` gets this value regardless of whether `max(d1,d2) <= 1e-15` or not.

src/ODE.jl
```diff
@@ -18,6 +18,41 @@ export ode4s, ode4ms, ode4
...
+        h0 = 1e-6
+	else
+        h0 = 1e-2d0/d1
+	end
+    # perform Euler step
+    x1 = x0 + h0*f0
+    f1 = F(t0 + h0, x1)
+    # estimate second derivative
+    d2 = norm(f1 - f0, Inf)/(tau*h0)
+    if max(d1, d2) <= 1e-15
```
tlycken: I haven't studied the algorithm, but is `1e-15` here really intended to be equivalent to `eps()`? If so, we might want to explicitly say `eps(<some quantity>)` or at least `eps(typeof(<some quantity>))` to stay generic (what if the user wants to solve stuff with `Float32`, or `BigFloat`?).

src/ODE.jl
```diff
@@ -456,9 +506,14 @@ function ode23s(F, y0, tspan; reltol = 1.0e-5, abstol = 1.0e-8,
     tfinal = tspan[end]
     tdir = sign(tfinal - t)
 
-    hmax = abs(tfinal - t)/10
-    hmin = abs(tfinal - t)/1e18
-    h = tdir*abs(tfinal - t)/100  # initial step size
+    hmax = maxstep
```
tlycken: Is there a reason not to just use the original variables? Why do we need to alias them?

tlycken:
Sorry. I took a look at the code and commented on a few nitpicks, but I haven't tried to run anything (I don't work with related stuff anymore, so I can do very little except look at the test results).

The CI build seems fine - Coveralls doesn't seem to be working on 0.4, but all the tests pass on both 0.3 and 0.4. No idea why Github didn't get a notification that the build was finished, but it's probably not your fault :)

tlycken:
And as far as the intent of the PR is concerned, I'm all for it :)

pwl:
Regarding the initial step choice, I also think you should replace the arbitrary constants (1e-5, 1e-15 etc.) with `eps(...)`.  For example, you could replace 1e-5 with `eps(...)^(1/3)`.  Although, I have no idea how the algorithm works and which constants should depend on the available precision.

By the way, if we want to have generic solvers, shouldn't we replace the numbers in Butcher tables with rationals?

tlycken:
@pwl regarding rationals: yes, we should.

src/ODE.jl
acroy: In the book they use those values, but don't say anything about the FP precision. But I guess your suggestion makes sense.

acroy:
I think the constants in `hinit` are just empirical values, which have proven to work well (probably for double precision computations). We can try to replace them, but it might be problematic.

tlycken:
Ah. Well, that does make it a little less obvious what to do, but I still think there's merit in trying. For most cases, it's quite obvious if the constant could chosen due to its relation to `eps(Float64)` or not (e.g. always splitting the time domain into 100 pieces is probably not related to floating point precision, while `1e-5 ~ cbrt(eps(Float64))` might be). I think we could try to use `eps()` whenever there's a straightforward way to get to the same value from it, and use `convert()` on the literal in all other cases to avoid promotion.

This might be better handled under #60 though, so I don't think it's necessary to change that before this PR can be merged. (Please do fix the indentation, though :) )

dpsanders:
Just a quick comment -- does anyone solve ODEs with rationals? (I guess not...) `eps` doesn't make sense in that context. If we're only talking about floating point, then putting the constants in terms of `eps` sounds like a good solution.

tlycken:
@acroy Are you using some IDE to do the indentation for you? There seems to be whitespace changes in a whole bunch of irrelevant places now, and the indentation is still off in many places (e.g. in `hinit` it seems that some rows have a tab size of 8 spaces and some (the `if` statements) have a tab size of 4)...

acroy:
@tlycken I am using Atom as an editor. I thought that I might have (accidentally) converted the tabs to spaces at some point. Strange.

acroy:
Now it should be better. I also changed `hinit` to return the value of `F` at the initial time, which saves us one evaluation.

tlycken:
Yeah, this LGTM.

acroy:
Should we use the opportunity to replace `ode23` by `ode23_bs`? I haven't updated the former since we wanted to replace it anyways.

ViralBShah:
Perhaps pick a better name, if possible?

acroy:
You mean `_bs` doesn't sound serious? :-) However, we would only use it internally and still export the name `ode23` (which calls `ode23_bs`) like we do it for `ode45` (which calls `ode45_dp`).

ViralBShah:
I think it sounds too serious, but is difficult for a newcomer to decipher, if it was a public api. :-) Internal API, of course, is ok.

tlycken:
I'd say it's time, but merge them separately to make it easier to revert one but not the other, should we need to.

acroy:
Alright, let's do it like that.

Issue #26, opened 2014-03-07
What to do with integer inputs?


ivarne:
A common dilemma in Julia, is what to do when the user provides integers to a function where they probably want do do the calculation in floating point. Many users do not care about the difference between `13.` and `13`, and does not want the "wrong" result when they forget the `.`. 

Suggestions
1. Try to solve the problem in integer space, and fail with `InexactError` 99.9 % of the time
2. Convert to `float` internally, but convert back to Int before returning
3. Error
4. Silently convert `T <: Integer` to `float`, and return floating point results.

It seems to be the general opinion that ODE solvers are performance sensitive, and we want them to be type stable, both internally and externally, to improve performance. We then have to decide what we want to do when the user supplies integers to `tspan` and `y0`.

ChrisRackauckas:
>There is no reasonable way to use integer or rational values in a numerical ODE solver except by promoting them to floating point.

Yes there is. With fixed timestep methods this is clearly possible. With adaptive methods, you just have to check your timestepping to change by rational values. One scheme for this is do reject/accept by doubling/halving the timestep.

Using integer and rational values for time can be good for ODE solvers which form the backend to DDE solvers since those are not very robust to small intervals that are due to propagation of discontinuities (this arises since 1/3 + 1/3 + 1/3 != 1 in floating point).

PR #60, opened 2015-03-26
WIP: Make coefficients Rational for better generics

tlycken:
*This is still just WIP, but I need some advice to keep this moving.*

@pwl [mentioned](https://github.com/JuliaLang/ODE.jl/pull/59#issuecomment-86431612) that it might be a good idea to convert all the coefficients in this code to rational, to make generic programming easier. I agree, since `Float64` is quite poisonous in this aspect; the promotion of almost anything and a `Float64` will be a `Float64`. We've [discussed before](https://github.com/JuliaLang/ODE.jl/issues/26) how to handle inputs of types that don't make sense in an ODE context and agreed that converting to floats is the most rational (heh...) thing to do, but there still might, for example, be applications for `Float32` arguments, or arguments of some user-defined number type, where we don't want to force `Float64`s on the user.

The PR currently only converts the coefficient tables that were already specified as fractions, but there is [one table that wasn't](https://github.com/JuliaLang/ODE.jl/blob/4a71fa0d97011aa4a35244e4685693bb99292e10/src/ODE.jl#L589-L598) (and I couldn't find a reference where it was) and there's also a bunch of coefficients and constants scattered throughout the various solvers, which are all specified as `Float64` literals. Do we want to try to make something more sensible with those? What?

It could also be argued that making the code generic yet type stable is too complicated to be worth the effort, in which case it's a good thing that I didn't spend more than a few minutes on this PR yet. (But I'd hate to end the discussion about this with "a `Float64` ought to be enough for anyone"...)


tlycken:
(The build failures are because Coveralls doesn't work on 0.4 - all our tests pass on both versions...)

mauro3:
The original paper by Kaps and Rentrop (1979) doesn't give ratios for the GRK4T coefficients, so presumably they are not know/don't exist.

stevengj:
One problem is that conversion of constant rational coefficients to floating-point cannot be done at compile time, so we have to be cautious about using these in inner loops.

mauro3:
Could/should `MathConst` be used instead?

Probably better convert the table before the inner loops to the used float-type.

mauro3:
I think this has been addressed by #68, right?

tlycken:
So it would seem.

PR #61, opened 2015-03-29
Change ode23 to call oderkf

acroy:
As discussed this changes `ode23` to call `oderkf` with Bogacki–Shampine coefficients. The original `ode23` is renamed to `ode23_ncm` (and moved to a separate file) for comparison and regression detection. README and LICENSE are updated.


tlycken:
LGTM, although I'm not sure why we want to keep and export the old implementation - aren't we running the same tests on the new one as we did on the old one?

Btw, the Travis build failed once due to connectivity issues with Github, probably related to the ongoing DDoS attack against GH. I restarted it hoping for better luck this time.

src/ode23_ncm.jl
```diff
@@ -0,0 +1,102 @@
...
+#   The initial condition is y(T0) = Y0.
+#
+#   The first argument, F, is a function handle or an anonymous function
+#   that defines f(t,y).  This function must have two input arguments,
+#   t and y, and must return a column vector of the derivatives, dy/dt.
+#
+#   With two output arguments, [T,Y] = ODE23(...) returns a column
+#   vector T and an array Y where Y(:,k) is the solution at T(k).
+#
+#   More than four input arguments, ODE23(F,TSPAN,Y0,RTOL,P1,P2,...),
```
mauro3: This paragraph seems wrong, no?

acroy:
It's not exported. But I also don't mind to remove it. 

I also had problems to push to my github repo. Let's hope they sort it out.

src/ode23_ncm.jl
acroy: Yes, it is outdated. Part of #57 was about fixing this.

acroy:
Thanks to @tlycken's patch I hope that both Travis runs will pass now. The question remains if we want to keep the old `ode23` or not?

tlycken:
Is the new implementation independent enough that we can drop any possible licensing issues we had with the old one? (I seem to recall a discussion about MATLAB's license not being MIT-compatible...)

acroy:
I don't know about the form of license of the old `ode23` (in LICENSE.md it only says "Copyright 2012 Cleve Moler and The MathWorks, Inc."). Our `oderkf` probably counts as "derived work" from Octave's `ode45`, which is GPL. I think it wouldn't be difficult to reimplement it (say using Hairer's book), but obviously (?) I can't do that. 
For`ode23s` I only used the paper of Shampine & Reichelt and I put it under MIT license, which is hopefully OK.

mauro3:
I that case I'd say, better remove `ode23`.  I made an explicit RK implementation following Hairer's book.  Once I've cleaned it up, I'll post it, maybe it can replace `oderkf`.

tlycken:
@mauro3 Nice!

@acroy To me it sounds like we don't need the old `ode23` for anything anymore, so you can delete it if you like, or merge this PR as is and we'll delete it together with the current `oderkf`. I'm fine either way.

acroy:
I removed `ode23_ncm` in the second commit.

@mauro3 : It would be great to have your RK implementation in ODE.jl. And it shouldn't be too hard to replace `oderkf`.

ViralBShah:
I had taken `ode23` was from Cleve Moler's textbook, and `ode45` from octave originally. So the comments about the licenses of both are spot on.

Perhaps it is best to remove `ode23`, and while GPL is perfectly fine, a non-GPL ode solver if @mauro3 has one, is also nice to have.

tlycken:
@acroy Are you waiting for something specific to merge this?

acroy:
No, it was just Easter dementia :-)

PR #62, opened 2015-04-02
Move coverage stat submission to after_success

tlycken:
As indicated in https://github.com/IainNZ/Coverage.jl/issues/54#issuecomment-87842579.

This might not be enough to fix the build script, but I don't know of any way to test it other than to push a PR.


tlycken:
Unclear if it was this or JakeBolewski/JuliaParser.jl#18 that broke it, but it seems to be fixed now.

PR #67, opened 2015-04-14
Make hinit aware of the integration direction. Fixes #66.

### LINKED ISSUE #66: ode45 tries to evaluate function outside of provided range
This seems to be a new bug introduced by a recent commit, since I have code that used to work a few days ago and now throws an exception. I managed to narrow down the problem to the following.

Consider solving for `F(x)=1-(1-x)^1.5` on [0,1] with boundary condition `F(1)=1`. To avoid numerical problems at x=1, replace 1 with 0.99:
```julia
using ODE
diff_eq(x,F) = try
                 1.5*(1-x).^(.5)
               catch err
                 @show x
                 throw(err)
               end
ode45(diff_eq, .99, [.99,0.0])
```
This gives this output
```
x => 1.056
ERROR: DomainError
 in diff_eq at none:5
 in hinit at /home/niedermayer/.julia/v0.3/ODE/src/ODE.jl:39
 in oderkf at /home/niedermayer/.julia/v0.3/ODE/src/ODE.jl:85
 in ode45_dp at /home/niedermayer/.julia/v0.3/ODE/src/ODE.jl:190
```
The problem seems to be that `ode45` tries to evaluate the function at x=1.056, which is outside of the range [0,0.99].

acroy:
I added an argument `tdir` to `hinit` such that the latter knows about the integration direction. In one of the tests (dy/dt=y) I reversed `tspan`. Our adaptive solvers pass this test. However, solvers related to `oderosenbrock` (`ode4s*`) seem to have problems with this for some reason.

@pao: I sort of remember that you wrote the initial implementation. Any idea how to fix this?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -19,6 +19,7 @@ solvers = [
     ode5ms,
     ODE.ode4s_s,
     ODE.ode4s_kr,
+
     ODE.ode78_fb]
 
 for solver in solvers
@@ -41,6 +42,9 @@ for solver in solvers
     t,y=solver((t,y)->y, 1., [0:.001:1;])
     @test maximum(abs(y-e.^t)) < tol
 
+    t,y=solver((t,y)->y, 1., [1:-.001:0;])
+    @test maximum(abs(y-e.^(t-1))) < tol
+ 
     # dv       dw
     # -- = -w, -- = v ==> v = v0*cos(t) - w0*sin(t), w = w0*cos(t) + v0*sin(t)
     # dt       dt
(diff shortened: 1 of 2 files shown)


acroy:
Never mind, I think I fixed it.

test/runtests.jl
```diff
@@ -19,6 +19,7 @@ solvers = [
     ode5ms,
     ODE.ode4s_s,
     ODE.ode4s_kr,
+
```
tlycken: ? :smile:

tlycken:
Nice catch on `oderosenbrok`. This LGTM :)

Issue #37, opened 2014-06-14
API clarification


berceanu:
I would like to solve the 2 coupled time-dependent PDEs from http://scicomp.stackexchange.com/questions/11772/numerical-approach-to-coupled-nonlinear-pdes-with-time-dependence/11782?noredirect=1#comment18452_11782
using ode45. I already have a working Fortran 90 code for this, based on the ode45 presented in Numerical Recipes. The plan is to add this to the examples section once finished.
I find the last test in https://github.com/JuliaLang/ODE.jl/blob/master/test/runtests.jl to be a good starting point, as it has a system of 2 coupled equations. However, my case is more complicated because:
1. the unknown functions `v` and `w` (called psiX and psiC in my problem) are themselves (complex) matrices, defined on a finite grid.
2. the RHS function (t, y) -> f is more involved, as it depends on a number of constants like kp, omegap, and the matrix F(x,y)
3. the laplacian term \delta psiC on the RHS of the second eq. has to be taken care of spectrally, i.e. fourier transforming, multipliying by k^2 and transforming back.

The main issue here: is it possible to use the current API to implement this problem? If not, I am willing to contribute in improving the API, but would appreciate some guidance.

Issue #20, opened 2014-03-05
API for ODE solvers


acroy:
This is a summary of the discussions in #7 and #4 and intended to reflect the latest decisions about the API.

Each (adaptive) solver accepts 3 arguments (PR in #14 )
- `F`: the RHS of the ODE `dy/dt = F(t,y)`, which is a function of `t` and `y(t)` and returns `dy/dt::typeof(y)`
- `y0`: initial value (no type specified -> duck typing)
- `tspan`: an array of points at which the solution is requested (`length(tspan)>=2`), the elements can be ordered ascending or descending (see #2)

Moreover, the following keywords are supported
- `norm`: user-supplied norm for determining the error `E` (default `Base.norm`)
- `abstol` and/or `reltol`: an integration step is accepted if `E <= abstol || E <= reltol*abs(y)` (ideally we want both criteria for all solvers, **done** in #13)
- `points=:all | :specified | :final`: controls the type of output according to
  - `points==:all` (default) output is given for each value in `tspan` as well as for each intermediate point the solver used. 
  - `points==:specified` output is given only for each value in `tspan`.
  - `points==:final` output is given only for the last value in `tspan`, i.e. at the end of integration (**speculative**).
- `maxstep`, `minstep` and `initstep`: determine the maximal, minimal and initial integration step.

The solver returns two arrays (PR in #14)
- `tout`: points at which solutions were obtained (also see keyword `points`)
- `yout`: solutions at times `tout`; if `points=:all | :specified` is used, `yout::Vector{typeof(y0)}`. For
  `points==:final` the solution at the end of the integration `yout::typeof(y0)` is returned (**speculative**).

If we we decide to support non-adaptive solvers (see #9), those should have the same arguments and return values.

PR #68, opened 2015-04-20
RFC: new adaptive (and fixed step) Runge Kutta solver

### LINKED ISSUE #25: Possibly implement dopri5

mauro3:
As promised in #61, here a RK implementation sans license restrictions.   It is a bit longer than the current code but has some more features as well.  ~~There is probably a bug lurking somewhere as it is less accurate than the current solvers.~~

Extra features:
- Tableaus are a datatype.  This would allow dispatch on different methods.  ~~Also, this "solves" what was tried in PR #60: the method will run using the type of number used for the coefficients.~~
- dense output (using 3rd order polynomials)

Improvement in ODE.jl to have `hinit` return `tdir` as well.

TODO:
- [X] figure out why less accurate than current solvers.
- [X] run through IVPtestsuite.jl
- [X] ~~doc update~~ (there is no doc yet)
- [X] fix performance issues
- [X] update to conform to internal of other ODE solvers
- [x] remove old solvers
- [X] squash commits

This touches on issue #9 and fixes #25 (although that was fixed already, no?).

Extra features, maybe in this PR, maybe later:
- [X] Tableaus in terms of Rational` and automatic conversion to suitable type
- ~~iterator protocol~~
- ~~maybe implement dense output in terms of `Polynomials.jl`,  root finding.~~

### DIFF
--- a/src/runge_kutta.jl
+++ b/src/runge_kutta.jl
@@ -1,0 +1,480 @@
+# Explicit Runge-Kutta solvers
+##############################
+# (Hairer & Wanner 1992 p.134, p.165-169)
+
+###########################################
+# Tableaus for explicit Runge-Kutta methods
+###########################################
+
+immutable TableauRKExplicit{Name, S, T} <: Tableau{Name, S, T}
+    order::(@compat(Tuple{Vararg{Int}})) # the order of the methods
+    a::Matrix{T}
+    # one or several row vectors.  First row is used for the step,
+    # second for error calc.
+    b::Matrix{T}
+    c::Vector{T}
+    function TableauRKExplicit(order,a,b,c)
+        @assert isa(S,Integer)
+        @assert isa(Name,Symbol)
+        @assert c[1]==0
+        @assert istril(a)
+        @assert S==length(c)==size(a,1)==size(a,2)==size(b,2)
+        @assert size(b,1)==length(order)
+        @assert norm(sum(a,2)-c'',Inf)<1e-10 # consistency.  
+        new(order,a,b,c)
+    end
+end
+function TableauRKExplicit{T}(name::Symbol, order::(@compat(Tuple{Vararg{Int}})),
+                   a::Matrix{T}, b::Matrix{T}, c::Vector{T})
+    TableauRKExplicit{name,length(c),T}(order, a, b, c)
+end
+function TableauRKExplicit(name::Symbol, order::(@compat(Tuple{Vararg{Int}})), T::Type,
+                   a::Matrix, b::Matrix, c::Vector)
+    TableauRKExplicit{name,length(c),T}(order, convert(Matrix{T},a),
+                                        convert(Matrix{T},b), convert(Vector{T},c) )
+end
+conv_field{T,N}(D,a::Array{T,N}) = convert(Array{D,N}, a)
+function Base.convert{Tnew<:Real,Name,S,T}(::Type{Tnew}, tab::TableauRKExplicit{Name,S,T})
+    # Converts the tableau coefficients to the new type Tnew
+    newflds = ()
+    @compat for n in fieldnames(tab)
+        fld = getfield(tab,n)
+        if eltype(fld)==T
+            newflds = tuple(newflds..., conv_field(Tnew, fld))
+        else
+            newflds = tuple(newflds..., fld)
+        end
+    end
+    TableauRKExplicit{Name,S,Tnew}(newflds...) # TODO: could this be done more generically in a type-stable way?
+end
+
+
+isexplicit(b::TableauRKExplicit) = istril(b.a) # Test whether it's an explicit method
+isadaptive(b::TableauRKExplicit) = size(b.b, 1)==2
+
+
+# First same as last.  Means ks[:,end]=ks_nextstep[:,1], c.f. H&W p.167
+isFSAL(btab::TableauRKExplicit) = btab.a[end,:]==btab.b[1,:] && btab.c[end]==1 # the latter is not needed really
+
+## Tableaus for explicit RK methods
+# Fixed step:
+const bt_feuler = TableauRKExplicit(:feuler,(1,), Rational{Int64},
+                                    zeros(Int,1,1),
+                                    [1]',
+                                    [0]
+                                    )
+const bt_midpoint = TableauRKExplicit(:midpoint,(2,), Rational{Int64},
+                                      [0  0
+                                       1//2  0],
+                                      [0, 1]',
+                                      [0, 1//2]
+                                      )
+const bt_heun = TableauRKExplicit(:heun,(2,), Rational{Int64},
+                                  [0  0
+                                   1  0],
+                                  [1//2, 1//2]',
+                                  [0, 1])
+
+const bt_rk4 = TableauRKExplicit(:rk4,(4,),Rational{Int64},
+                                 [0    0    0 0
+                                  1//2 0    0 0
+                                  0    1//2 0 0
+                                  0    0    1 0],
+                                 [1//6, 1//3, 1//3, 1//6]',
+                                 [0, 1//2, 1//2, 1])
+
+# Adaptive step:
+# Heun Euler https://en.wikipedia.org/wiki/Runge–Kutta_methods
+const bt_rk21 = TableauRKExplicit(:heun_euler,(2,1), Rational{Int64},
+                                  [0     0
+                                   1     0],
+                                  [1//2  1//2
+                                   1     0],
+                                  [0,    1])
+
+# Bogacki–Shampine coefficients
+const bt_rk23 = TableauRKExplicit(:bogacki_shampine,(2,3), Rational{Int64},
+                                  [0           0      0      0
+                                   1/2         0      0      0
+                                   0         3/4      0      0
+                                   2/9       1/3     4/9     0],
+                                  [7/24 1/4 1/3 1/8
+                                   2/9 1/3 4/9 0],
+                                  [0, 1//2, 3//4, 1] 
+                         )
+
+# Fehlberg https://en.wikipedia.org/wiki/Runge%E2%80%93Kutta%E2%80%93Fehlberg_method
+const bt_rk45 = TableauRKExplicit(:fehlberg,(4,5),Rational{Int64},
+                          [  0           0           0            0         0     0
+                             1//4        0           0            0         0     0
+                             3//32       9//32       0            0         0     0
+                          1932//2197 -7200//2197  7296//2197      0         0     0
+                           439//216     -8        3680//513    -845//4104   0     0
+                            -8//27       2       -3544//2565   1859//4104 -11//40 0 ],
+                           [25//216      0        1408//2565   2197//4104  -1//5  0
+                            16//135      0        6656//12825 28561//56430 -9//50 2//55],
+                            [0,          1//4,       3//8,       12//13,    1,    1//2])
+
+# Dormand-Prince https://en.wikipedia.org/wiki/Dormand%E2%80%93Prince_method
+const bt_dopri5 = TableauRKExplicit(:dopri, (5,4), Rational{Int64},
+                     [0           0            0                   0            0      0 0
+                      1//5        0            0                   0            0      0 0
+                      3//40       9//40        0                   0            0      0 0
+                      44//45      -56//15      32//9               0            0      0 0
+                      19372//6561 -25360//2187 64448//6561 -212//729            0      0 0
+                      9017//3168  -355//33     46732//5247   49//176 -5103//18656      0 0
+                      35//384     0            500//1113    125//192  -2187//6784 11//84 0],
+                     [35//384         0     500//1113       125//192      -2187//6784         11//84      0
+                      5179//57600     0     7571//16695     393//640     -92097//339200     187//2100     1//40],
+                     [0, 1//5, 3//10, 4//5, 8//9, 1, 1]
+                     )
+
+# Fehlberg 7(8) coefficients
+# Values from pag. 65, Fehlberg, Erwin. "Classical fifth-, sixth-, seventh-, and eighth-order Runge-Kutta formulas with stepsize control".
+# National Aeronautics and Space Administration.
+const bt_feh78 = TableauRKExplicit(:feh78, (7,8), Rational{Int64},
+                            [     0       0      0       0         0          0        0        0      0       0     0 0 0
+                                  2//27   0      0       0         0          0        0        0      0       0     0 0 0
+                                  1//36   1//12  0       0         0          0        0        0      0       0     0 0 0
+                                  1//24   0      1//8    0         0          0        0        0      0       0     0 0 0
+                                  5//12   0    -25//16  25//16     0          0        0        0      0       0     0 0 0
+                                  1//20   0      0       1//4      1//5       0        0        0      0       0     0 0 0
+                                -25//108  0      0     125//108  -65//27    125//54    0        0      0       0     0 0 0
+                                 31//300  0      0       0       61//225    -2//9     13//900   0      0       0     0 0 0
+                                  2       0      0     -53//6    704//45   -107//9    67//90    3      0       0     0 0 0
+                                -91//108  0      0      23//108 -976//135   311//54  -19//60   17//6  -1//12   0     0 0 0
+                               2383//4100 0      0    -341//164 4496//1025 -301//82 2133//4100 45//82 45//164 18//41 0 0 0
+                                  3//205  0      0       0        0        -6//41     -3//205  -3//41  3//41   6//41 0 0 0
+                              -1777//4100 0      0    -341//164 4496//1025 -289//82 2193//4100 51//82 33//164 12//41 0 1 0],
+                              [41//840 0 0 0 0 34//105 9//35 9//35 9//280 9//280 41//840 0 0
+                               0 0 0 0 0 34//105 9//35 9//35 9//280 9//280 0     41//840 41//840],
+                               [0,    2//27, 1//9, 1//6 , 5//12, 1//2 , 5//6 , 1//6 , 2//3 , 1//3 , 1 , 0, 1]
+                            )
+
+
+################################
+# Fixed step Runge-Kutta methods
+################################
+
+# TODO: iterator method
+ode1(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_feuler)
+ode2_midpoint(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_midpoint)
+ode2_heun(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_heun)
+ode4(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_rk4)
+
+function oderk_fixed(fn, y0, tspan, btab::TableauRKExplicit)
+    # Non-arrays y0 treat as scalar
+    fn_(t, y) = [fn(t, y[1])]
+    t,y = oderk_fixed(fn_, [y0], tspan, btab)
+    return t, vcat_nosplat(y)
+end
+function oderk_fixed{N,S}(fn, y0::AbstractVector, tspan,
+                          btab_::TableauRKExplicit{N,S})
+    # TODO: instead of AbstractVector use a Holy-trait
+
+    # Needed interface:
+    # On components: 
+    # On y0 container: length, deepcopy, similar, setindex!
+    # On time container: getindex, convert. length
+
+    Et, Eyf, Ty, btab = make_consistent_types(fn, y0, tspan, btab_)
+    dof = length(y0)
+
+    ys = Array(Ty, length(tspan))
+    allocate!(ys, y0, dof)
+    ys[1] = deepcopy(y0)
+
+    tspan = convert(Vector{Et}, tspan)
+    # work arrays:
+    ks = Array(Ty, S)
+    # allocate!(ks, y0, dof) # no need to allocate as fn is not in-place
+    ytmp = similar(y0, Eyf, dof)
+    for i=1:length(tspan)-1
+        dt = tspan[i+1]-tspan[i]
+        ys[i+1][:] = ys[i]
+        for s=1:S
+            calc_next_k!(ks, ytmp, ys[i], s, fn, tspan[i], dt, dof, btab)
+            for d=1:dof
+                ys[i+1][d] += dt * btab.b[s]*ks[s][d]
+            end
+        end
+    end
+    return tspan, ys
+end
+
+##############################
+# Adaptive Runge-Kutta methods
+##############################
+
+ode21(fn, y0, tspan; kwargs...) = oderk_adapt(fn, y0, tspan, bt_rk21; kwargs...)
+ode23(fn, y0, tspan; kwargs...) = oderk_adapt(fn, y0, tspan, bt_rk23; kwargs...)
+ode45_fe(fn, y0, tspan; kwargs...) = oderk_adapt(fn, y0, tspan, bt_rk45; kwargs...)
+ode45_dp(fn, y0, tspan; kwargs...) = oderk_adapt(fn, y0, tspan, bt_dopri5; kwargs...)
+# Use Dormand-Prince version of ode45 by default
+const ode45 = ode45_dp
+ode78(fn, y0, tspan; kwargs...) = oderk_adapt(fn, y0, tspan, bt_feh78; kwargs...)
+
+function oderk_adapt(fn, y0, tspan, btab::TableauRKExplicit; kwords...)
+    # For y0 which don't support indexing.
+    fn_ = (t, y) -> [fn(t, y[1])]
+    t,y = oderk_adapt(fn_, [y0], tspan, btab; kwords...)
+    return t, vcat_nosplat(y)
+end
+function oderk_adapt{N,S}(fn, y0::AbstractVector, tspan, btab_::TableauRKExplicit{N,S};
+                          reltol = 1.0e-5, abstol = 1.0e-8,
+                          norm=Base.norm,
+                          minstep=abs(tspan[end] - tspan[1])/1e9,
+                          maxstep=abs(tspan[end] - tspan[1])/2.5,
+                          initstep=0.,
+                          points=:all
+                          )
+    # Needed interface:
+    # On components:
+    #  - note that the type of the components might change!
+    # On y0 container: length, similar, setindex!
+    # On time container: getindex, convert, length
+    
+    # For y0 which support indexing.  Currently y0<:AbstractVector but
+    # that could be relaxed with a Holy-trait.
+    !isadaptive(btab_) && error("Can only use this solver with an adaptive RK Butcher table")
+
+    Et, Eyf, Ty, btab = make_consistent_types(fn, y0, tspan, btab_)
+    # parameters
+    order = minimum(btab.order)
+    timeout_const = 5 # after step reduction do not increase step for
+                      # timeout_const steps
+
+    ## Initialization
+    dof = length(y0)
+    tspan = convert(Vector{Et}, tspan)
+    t = tspan[1]
+    tstart = tspan[1]
+    tend = tspan[end]
+
+    # work arrays:
+    y      = similar(y0, Eyf, dof)      # y at time t
+    y[:]   = y0
+    ytrial = similar(y0, Eyf, dof) # trial solution at time t+dt
+    yerr   = similar(y0, Eyf, dof) # error of trial solution
+    ks = Array(Ty, S)
+    # allocate!(ks, y0, dof) # no need to allocate as fn is not in-place
+    ytmp   = similar(y0, Eyf, dof)
+
+    # output ys
+    nsteps_fixed = length(tspan) # these are always output
+    ys = Array(Ty, nsteps_fixed)
+    allocate!(ys, y0, dof)
+    ys[1] = y0
+
+    # Option points determines where solution is returned:
+    if points==:all
+        tspan_fixed = tspan
+        tspan = Et[tstart]
+        iter_fixed = 2 # index into tspan_fixed
+        sizehint!(tspan, nsteps_fixed)
+    elseif points!=:specified
+        error("Unrecognized option points==$points")
+    end
+    # Time
+    dt, tdir, ks[1] = hinit(fn, y, tstart, tend, order, reltol, abstol) # sets ks[1]=f0
+    if initstep!=0
+        dt = sign(initstep)==tdir ? initstep : error("initstep has wrong sign.")
+    end
+    # Diagnostics
+    dts = Et[]
+    errs = Float64[]
+    steps = [0,0]  # [accepted, rejected]
+
+    ## Integration loop
+    laststep = false
+    timeout = 0 # for step-control
+    iter = 2 # the index into tspan and ys
+    while true
+        # do one step (assumes ks[1]==f0)
+        rk_embedded_step!(ytrial, yerr, ks, ytmp, y, fn, t, dt, dof, btab)
+        # Check error and find a new step size:
+        err, newdt, timeout = stepsize_hw92!(dt, tdir, y, ytrial, yerr, order, timeout,
+                                            dof, abstol, reltol, maxstep, norm)
+
+        if err<=1.0 # accept step
+            # diagnostics
+            steps[1] +=1
+            push!(dts, dt)
+            push!(errs, err)
+
+            # Output:
+            f0 = ks[1]
+            f1 = isFSAL(btab) ? ks[S] : fn(t+dt, ytrial)
+            if points==:specified
+                # interpolate onto given output points
+                while iter-1<nsteps_fixed && (tdir*tspan[iter]<tdir*(t+dt) || laststep) # output at all new times which are < t+dt
+                    hermite_interp!(ys[iter], tspan[iter], t, dt, y, ytrial, f0, f1) # TODO: 3rd order only!
+                    iter += 1
+                end
+            else
+                # first interpolate onto given output points
+                while iter_fixed-1<nsteps_fixed && tdir*t<tdir*tspan_fixed[iter_fixed]<tdir*(t+dt) # output at all new times which are < t+dt
+                    yout = hermite_interp(tspan_fixed[iter_fixed], t, dt, y, ytrial, f0, f1)
+                    index_or_push!(ys, iter, yout) # TODO: 3rd order only!
+                    push!(tspan, tspan_fixed[iter_fixed])
+                    iter_fixed += 1
+                    iter += 1
+                end
+                # but also output every step taken
+                index_or_push!(ys, iter, deepcopy(ytrial))
+                push!(tspan, t+dt)
+                iter += 1
+            end
+            ks[1] = f1 # load ks[1]==f0 for next step
+
+            # Break if this was the last step:
+            laststep && break
+
+            # Swap bindings of y and ytrial, avoids one copy
+            y, ytrial = ytrial, y
+
+            # Update t to the time at the end of current step:
+            t += dt
+            dt = newdt
+
+            # Hit end point exactly if next step within 1% of end:
+            if tdir*(t+dt*1.01) >= tdir*tend
+                dt = tend-t
+                laststep = true # next step is the last, if it succeeds
+            end
+        elseif abs(newdt)<minstep  # minimum step size reached, break
+            println("Warning: dt < minstep.  Stopping.")
+            break
+        else # redo step with smaller dt
+            laststep = false
+            steps[2] +=1
+            dt = newdt
+            timeout = timeout_const
+        end
+    end
+    return tspan, ys
+end
+
+function rk_embedded_step!{N,S}(ytrial, yerr, ks, ytmp, y, fn, t, dt, dof, btab::TableauRKExplicit{N,S})
+    # Does one embedded R-K step updating ytrial, yerr and ks.
+    #
+    # Assumes that ks[:,1] is already calculated!
+    #
+    # Modifies ytrial, yerr, ks, and ytmp
+
+    # Needed interface:
+    # On components: arithmetic, zero
+    # On y0 container: fill!, setindex!, getindex
+
+    fill!(ytrial, zero(eltype(ytrial)) )
+    fill!(yerr, zero(eltype(ytrial)) )
+    for d=1:dof
+        ytrial[d] += btab.b[1,1]*ks[1][d]
+        yerr[d]   += btab.b[2,1]*ks[1][d]
+    end
+    for s=2:S
+        calc_next_k!(ks, ytmp, y, s, fn, t, dt, dof, btab)
+        for d=1:dof
+            ytrial[d] += btab.b[1,s]*ks[s][d]
+            yerr[d]   += btab.b[2,s]*ks[s][d]
+        end
+    end
+    for d=1:dof
+        yerr[d]   = dt * (ytrial[d]-yerr[d])
+        ytrial[d] = y[d] + dt * ytrial[d]
+    end
+end
+
+function stepsize_hw92!(dt, tdir, x0, xtrial, xerr, order,
+                       timeout, dof, abstol, reltol, maxstep, norm)
+    # Estimates the error and a new step size following Hairer &
+    # Wanner 1992, p167 (with some modifications)
+    #
+    # If timeout>0 no step size increase is allowed, timeout is
+    # decremented in here.
+    #
+    # Returns the error, newdt and the number of timeout-steps
+    #
+    # TODO:
+    # - allow component-wise reltol and abstol?
+    # - allow other norms
+
+    # Needed interface:
+    # On components: isoutofdomain
+    # On y0 container: norm, get/setindex
+
+    timout_after_nan = 5
+    fac = [0.8, 0.9, 0.25^(1/(order+1)), 0.38^(1/(order+1))][1]
+    facmax = 5.0 # maximal step size increase. 1.5-5
+    facmin = 1./facmax  # maximal step size decrease. ?
+
+    # in-place calculate xerr./tol
+    for d=1:dof
+        # if outside of domain (usually NaN) then make step size smaller by maximum
+        isoutofdomain(xtrial[d]) && return 10., dt*facmin, timout_after_nan
+        xerr[d] = xerr[d]/(abstol + max(norm(x0[d]), norm(xtrial[d]))*reltol) # Eq 4.10
+    end
+    err = norm(xerr, 2) # Eq. 4.11
+    newdt = min(maxstep, tdir*dt*max(facmin, fac*(1/err)^(1/(order+1)))) # Eq 4.13 modified
+    if timeout>0
+        newdt = min(newdt, dt)
+        timeout -= 1
+    end
+    return err, tdir*newdt, timeout
+end
+
+function calc_next_k!{Ty}(ks::Vector, ytmp::Ty, y, s, fn, t, dt, dof, btab)
+    # Calculates the next ks and puts it into ks[s]
+    # - ks and ytmp are modified inside this function.
+
+    # Needed interface:
+    # On components: +, *
+    # On y0 container: setindex!, getindex, fn
+
+    ytmp[:] = y
+    for ss=1:s-1, d=1:dof
+        ytmp[d] += dt * ks[ss][d] * btab.a[s,ss]
+    end
+    ks[s] = fn(t + btab.c[s]*dt, ytmp)::Ty
+    nothing
+end
+
+# Helper functions:
+function allocate!{T}(vec::Vector{T}, y0, dof)
+    # Allocates all vectors inside a Vector{Vector} using the same
+    # kind of container as y0 has and element type eltype(eltype(vec)).
+    for s=1:length(vec)
+        vec[s] = similar(y0, eltype(T), dof)
+    end
+end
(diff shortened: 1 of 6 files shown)


src/runge_kutta.jl
```diff
@@ -0,0 +1,396 @@
...
+
+# Test whether it's an explicit method
+isexplicit(b::Tableau) = istril(b.a)
+isadaptive(b::Tableau) = size(b.b, 1)==2
+order(b::Tableau) = b.order
+
+# The advantage of each having its own type, makes it possible to have
+# specialized methods for a particular tablau
+immutable TableauRKExplicit{Name, S, T} <: Tableau{Name, S, T}
+    order::(Int...) # the order of the methods
```
tlycken: This will break on 0.4 with the newly introduced tuple changes; at the moment, it should probably be `order::(@compat(Tuple{Vararg{Int}}))` but I'm unsure on how this works in this context; the outer parenthesis might throw this off.

It will work on 0.3 as is, though :)

tlycken:
Actually, this doesn't solve the problem that #60 addresses, since you've effectively converted all the tables to `Float64` anyway in the constructors of `TableauRKExplicit` (when saving the arrays). It's a good way on its way, though, and the approach is better than what I had in #60, so I'd gladly yield that for this! :)

mauro3:
Ah, yes, you're right.  Another step is needed.  But my idea was that the parameter `T` of the tableau would determine what type is used in the solver.  As it doesn't make any sense to have the tabelau in a different type to the numerics.  So, we'd need to write it:
```julia
bt_rk45_coef = (          [  0           0           0            0         0     0
                             1//4        0           0            0         0     0
                             3//32       9//32       0            0         0     0
                          1932//2197 -7200//2197  7296//2197      0         0     0
                           439//216     -8        3680//513    -845//4104   0     0
                            -8//27       2       -3544//2565   1859//4104 -11//40 0 ],
                           [25//216      0        1408//2565   2197//4104  -1//5  0
                            16//135      0        6656//12825 28561//56430 -9//50 2//55],
                            [0,          1//4,       3//8,       12//13,    1,    1//2])

bt_rk45 = TableauRKExplicit(:fehlberg,(4,5),Float64, bt_rk45_coef...)
bt_rk45_f32 = TableauRKExplicit(:fehlberg,(4,5),Float32, bt_rk45_coef...)
```

src/runge_kutta.jl
mauro3: I sure failed on Travis!

acroy:
The problem is, that the type of the solution is determined by the type of the time-step (most likely float), the type of the tableau coefficients **and** the type of `F` (rhs of the ODE). All three are somehow multiplied to give the new solution. So I am not sure it makes much sense to store the coefficients as Rational ...

Nice work BTW!

mauro3:
My idea was that 99% of users will work with Float64.  For the other 1%, they can instantiate something like `bt_rk45_f32` where `T==Float32`. As my solver is written, then initial conditions and time vector are, if necessary, converted to that `T` (actually this should be tested).  It would be up to the user to make sure that there is no conversion or no loss in the conversion. So, this is not fully automatic but can be extended if needed.  Note that inside `bt_rk45`, which is used by default, the coefficients are stored as Float64.

And thanks. But, I just had a look at performance with IVPtestSuite and it's atrocious, so still lots to be done...

src/runge_kutta.jl
```diff
@@ -0,0 +1,396 @@
...
+    TableauRKExplicit{name,length(c),T}(order, convert(Matrix{T},a),
+                                        convert(Matrix{T},b), convert(Vector{T},c) )
+end
+
+# First same as last, c.f. H&W p.167
+isFSAL(btab::TableauRKExplicit) = btab.a[end,:]==btab.b[1,:] && btab.c[end]==1 # the latter is not needed really
+
+## Tableaus for explicit methods
+# Fixed step:
+bt_feuler = TableauRKExplicit(:feuler,(1,), Float64,
```
stevengj: `const`?

tlycken:
The reason to store the coefficients as rational is that they are the least invasive when promoting with other number types. Then, we can do something conceptually equivalent to 

```julia
TTime = eltype(tspan)
TRHS = typeof(F(0, #= whatever args we need =#))
TTableau = typeof(one(TTime) * one(TRHS) * one(Rational{Int})
```

This way, `TTableau` will automatically be `Float32` for a `tspan`/`F` combo which returns `Float32`, remain `Rational{Int}` for such inputs etc. `Interpolations.jl` does very similar things to handle typing of coefficients gracefully.

pwl:
Sorry for nitpicking but I am in the 1% not using `Float64`.  Is this new solver generic in the sense that it would be possible to do computations using `BigFloat` without loosing precision on the way (via e.g. multiplications by explicit `Float64` constants)?  So far it kind of looks like `Float64` is hard-wired to the coefficient tables now.  I could definitely live with initial conversions at the expense slower solver startup.

tlycken:
@mauro3 The reason this won't work is that it's impossible for us to prepare in this way for the case where someone has a self-created number type which doesn't promote cleanly with `Float64` - consider for example a type `HumongousFloat` that has 1024 bits of precision instead of `BigFloat`'s 256. We can't create versions of these tables for such types beforehand, but we *can* convert to these types on invocation. Storing the coefficients as rationals ensures that it's possible to convert to any concievable number type without loss of precision.

@pwl I don't think the current version of this PR works like that - and it might turn out to be out of scope for this PR to make it work like that. However, I do think it's valuable enough that we should solve this at some point. We'll see where this leads to and take it from there, I guess.

However important generics and sensible promotion rules are, I think the main merit of this PR is to provide a white-room Runge-Kutta implementation, allowing us to greatly simplify the license of this package. That's no small thing, indeed! :)

pwl:
I just wanted to ensure that the 1% is well represented. :)  And I already have my own working fork so there is no hurry in implementing this feature.  Don't get me wrong, I see this PR as a remarkable piece of work.

mauro3:
After fixing a bug in the step control, this now works.  Here the performance tests from IVPTestSuites.jl:

scd-vs-walltime-threebody

The new solvers (`*_v2`) about a factor 2 faster, memory footprint is also down by a factor 2-5 (not shown).  However the slope of the old `78` solver is slightly shallower.

scd-vs-walltime-hires

All solvers new solvers a factor 2-5 faster.

And for the fixed step solver:

fixedstep-scd-vs-walltime-threebody

The solver is significantly better than the old ones.  This is a bit surprising as the `ode4` and `ode4_v2` should be the same alogrithm?!

mauro3:
@tlycken, yes I agree, that was the idea behind the little code example above, in `bt_rk45_coef` are the rational coefficients.  But maybe we should discuss this in another PR.

acroy:
Very nice :-)

>The solver is significantly better than the old ones. This is a bit surprising as the ode4 and ode4_v2 should be the same alogrithm?!

Well, in the old code many copies of the solution are made, maybe this explains the difference? Partly, this was necessary to make the code work with user-defined types.  At the moment, `y` doesn't have to be `<:AbstractArray` for example. Is this possible with your code as well?

src/runge_kutta.jl
```diff
@@ -0,0 +1,400 @@
...
+end
+
+# Fixed step Runge-Kutta method
+# TODO: iterator method
+function oderk_fixed{N,S,T}(fn, y0, tspan::AbstractVector,
+                            btab::TableauRKExplicit{N,S,T})
+    dof = length(y0)
+    tsteps = length(tspan)
+    ys = Array(T, dof, tsteps)
+    ys[:,1] = y0'
```
acroy: Its probably better to use `transpose` here.

mauro3:
@acroy: we should add tests for non-abstractarrays if we want to support them.  But what is the use case?

src/runge_kutta.jl
mauro3: Changed to `.'` (I think `.'` is sugar for `transpose`, right?)

mauro3:
@acroy: and no, at the moment it doesn't work quite work for those general cases.  It also stores the ys in a Matrix internally.  I can update that.

Related, I find the current interface returning a Vector of something quite annoying.  I can see that for the general case that is needed. But I always have to `hcat(y...)`.  Maybe return a Matrix for the normal case and only the Vector of Vectors otherwise?

mauro3:
@acroy: before updating the PR to allow for more general datatypes, a few questions:

What is the "interface" a `typeof(y)` has to support? Looking through the code, it looks like just arithmetic.  However, considering the speed and memory improvements provided by looping and indexing, it seems a shame to let those improvements slip.  So, could I use indexing?  Presumably it would be ok to use arrays for intermediate storage?

acroy:
I agree that the typical use case will involve scalars or dense vectors, but there might be good reasons to have a user-defined type, which doesn't support the full array interface. For example, in JuliaQuantum/QuBase.jl we have `QuArrays` which are not `<:AbstractArray`. Also AFAIK @Jutho's tensors are not abstract arrays. Nevertheless, we definitely want to use ODE.jl to solve some differential equations.

> Related, I find the current interface returning a Vector of something quite annoying.

Yeah, we discussed this when we put together the [API specification](https://github.com/JuliaLang/ODE.jl/blob/master/doc/api.md) (probably in #20). I think one point was, that one wants to use `yout[end]` as new input or to evaluate `F`. We could handle the array-input case separately, but this would also break the API consistency.

acroy:
@mauro: I think it is just `norm`, `+`, `-`, multiplication with/division by scalars. At some point (#37) I came up with an [example gist](https://gist.github.com/acroy/28be4f2384d01f38e577) to demonstrate the basics. But I agree that looping and indexing is really helpful, so maybe we should revise the requirements?

tlycken:
I haven't done any profiling, but if memory management is the bottleneck here, we could very probably use the iterator versions (#49) to let the user take care of it in cases where they want to:

* We keep `tout, yout = ode45(F, y0, tspan)` returning a `Vector{typeof(y0)}`
* If that is too slow for the user, they can pre-allocate space for their results in a matrix, and use the iterator interface to fill it.
* Internally, we use the iterator interface to fill the `Vector`, so we don't duplicate code

If we choose that path (and there is a clean-enough way to do it for the user, so we can show it off in examples), we can move that issue out of scope for this PR :)

mauro3:
I'm trying to adapt the solvers to the preferred API and internals but I find it quite painful and I'm wondering whether this in not a case of over-engineering? There are a few types involved:
```
Tt = typeof(tspan)
Ty = typeof(y0)
Tf = typeof(F(tspan(1),y0))

Et = eltype(tspan)
Ey = eltype(y0)
Ef = eltype(Tf)
```

`E*`: There should be promotion of the eltypes (#26,): Tt should probably be some type of floating point number, for sure a Real (maybe FixedPoint, Rational could maybe work but as @stevengj said, it doesn't make sense), Ty should be some "continuous" Number, Tf also.  

`T*`: The `T*` are either equal to the eltypes, if scalar, or some kind of container type. In the current solvers Tt is discarded, the outputs are kept in a `Vector{Ty}` (this is where I experience my pain) and it is assumed `Tf==Ty` (and thus `Ef==Ey`).

Anyway, I wonder whether there is any need to support other containers than AbstractArrays?  And in particular whether internal, temporary storage arrays can be just `Array`s or not.  If not, what is the use case?

The reason I am wondering these things is that I would like to see an API developed, similar to Sundials, where all storage can be preallocated and functions do in-place updated.  I don't see how this is compatible with above internals.  I don't think an iterator method would work, at least not if it is not in-place.

Sorry for the rant.  I'll ponder this a bit longer...

pwl:
The assumption that `Tf==Ty` is sensible, we always need as many equations as there are variables so it should be always possible to map the return value `Tf` back into `Ty`.  As for `Tt` I think it should be an (abstract) vector of Real type (interpreted as either `[start,stop]` or a sequence of predetermined output times).  Can you see any other uses for `Tt`?

On the other hand the in-place updates are a must have feature if we want good performance in the iterator version and I don't see how this could be realized with custom types of `Ty` and `Tf`.  Maybe by defining an indexing methods on `Ty` and preallocating instances of `Ty` as "working arrays"?  Would that be feasible?  Can we simply go with Arrays for this PR and think of this as a separate feature (as @tlycken suggests) or does it have to be a part of this PR for some technical reasons?

pwl:
And just a comment, in implicit methods `Ty` can only be an (abstract) vector because we have to solve an implicit equation for the next step via Newton iterations, which require a Jacobian for the whole system to be a 2d array.  In principle it might be possible to prepare a specialized array type for each call but I guess it won't be very efficient.

My opinion is that types other than vectors would be very impractical to work with inside ODE.jl for reasons already mentioned and I think we should refrain from implementing API this way.  Anyway, with iterators in place one can easily produce wrappers to generate user defined types from the array returned at each step via `imap` for example, same goes for the wrapper around `F` (I am doing this in my code and it works well enough).  If somebody wants to define a specialized type he can also create these wrappers without much effort.

Jutho:
I haven't followed this discussion, but the tensors in TensorToolbox.jl (which is far from usable and is currently receiving little attention) are indeed not `AbstractArray` subtypes. However, they are easily transformed into a `Vector` and back, which I think is the only way they can be used together with e.g. iterative solvers, eigensolvers, ode integrators, etc... I think of these tensors as including both the coefficients and information about the basis vectors that span the vector space (probably very close to how QuBase.jl is working). The coefficients itself can easily be returned as a `Vector`.

mauro3:
@Jutho I think, the normal use case involving TensorToolbox.jl would be to have a tensor field and time step that forward.  So one tensor would correspond to a "Number" and the field would be contained in a Array-like container (i.e. the tensor would be `Ey`, in above parlance).  Is that correct, or is there a tensor-field type as well?

@acroy Whereas, presumably, a QuArray would itself be the container (`Ty`), right?

@pwl: couldn't a jacobian be more abstract than a matrix? 

If, as @pwl suggests, a Array-version should be merged that would be commit 8ac21e25254fbd28537ffd4fd0a160da28bd6704.  Let me know if I should revert to that.  Otherwise, I'm not sure that I want to try and update the PR. At least not until a more satisfying solution to above points has been found.

acroy:
@mauro3 : That is right, `QuArray` is basically the container in that case. Of course we can also convert this to a vector and back if necessary.

Maybe I am the only one who wants to keep the support for more general types via duck typing... I can see that this might be painful, since things like `+=` make copies and we don't have canonical in-place operations like `add!` yet. However, it seems that circumventing this limitation/bottleneck is the only reason for restricting the input to arrays. (I wish I could give a killer-example, but I don't have one. I just think there is no fundamental reason why the user has to represent his/her system of equations by an array, which in the end might be the most efficient way, but that is a different problem). 

I hope this didn't sound too harsh, it certainly isn't meant to be. If everyone is in favor of switching to arrays, I am fine with that.

tlycken:
I am also strongly in favor of keeping the typing as open as possible via duck typing (although I don't necessarily think it has to be done in this PR).

My reasoning is this:

When it comes to "problems" like `+=` making copies, I think this is really something we should leave to base Julia to optimize. Given the tremendous impact a performance boost for such operations will give across the board, it's very unlikely that the core team won't have a solution in place well before Julia 1.0 is released.

If it turns out that there are users who really do need the extra performance gain right now, we could possibly write specialized methods for `Array` inputs that have these optimizations, but I think it's premature to have *only* those, and disallow a lot of other potential use cases.

mauro3:
My proposal is the following (which is implemented in the last commit b8458df):
```julia
# For non-arrays y0, treat y0 as scalar
function oderk_fixed(fn, y0, tspan, btab::TableauRKExplicit)
    fn_ = (t, y) -> fn(t, y[1])
    t,y = oderk_fixed(fn_, [y0], tspan, btab)
    return t,vcat(y...)
end
# For array-like y0 use this.  Note that the y0<:AbstractArray could be relaxed to 
# any indexing container with a Holy-trait
function oderk_fixed{N,S}(fn, y0::AbstractVector, tspan,
                          btab_::TableauRKExplicit{N,S})
    Ar, Et, Ey, Ef, btab = make_consistent_types(fn, y0, tspan, btab_)
    
    dof = length(y0)
    tsteps = length(tspan)
    ys = Ar(Ey, dof, tsteps)
    ys[:,1] = y0
    tspan = convert(Vector{Et}, tspan)
    # work arrays:
    ks = Ar(Ef, dof, S)
    ytmp = Ar(Ey, dof)

    for i=1:length(tspan)-1
        dt = tspan[i+1]-tspan[i]
        ys[:,i+1] = ys[:,i] # this also allocates a bit...
        for s=1:S
            for d=1:dof
                ytmp[d] = ys[d,i] # ytmp[:] = ys[:,i] allocates!
            end
            calc_next_k!(ks, ytmp, ytmp, s, fn, tspan[i], dt, dof, btab)
            for d=1:dof
                ys[d,i+1] += dt * btab.b[s]*ks[d,s]
            end
        end
    end
    return tspan, ys
end
```

I think this should work for arbitrary `y0` which cannot be indexed (need to test this), and works fast for indexable `y0`.  Note that `Ar` is a suitable container type which `make_consistent_types` figures out.

mauro3:
I updated the code slightly (but still a bit hacky at the moment) and also @acroy cutstom-type example:
https://gist.github.com/mauro3/d605bff8e11994a372d9
Now this works with both the adaptive stepper and fixed step.  So, my approach seems feasible.

mauro3:
My use case is: I want to use ode time steppers in the near future (next two month) to time-step discretized PDEs with up to 10^5 degrees of freedom.  (Currently I do this with Matlab and it works up to 2e4 DOFs). Whilst Julia may eventually make many of the bottlenecks go away, it will take some time to get there, certainly too long for me to wait.  Also, for systems that size you want to work without allocations.

But I also see the need/use for "scalar" values, like in @acroy's gist. I think the approach I outlined above could serve both needs, albeit with a somewhat higher code complexity.  Performance-wise it would probably favor non-scalar types but I think that is ok, as those are typically the large systems.

tlycken:
OK, I see.

How about actually having two versions of the solver, with the choice between them being made on the types? Then one can explicitly dispatch on `AbstractVector` (or even `Vector`, if it makes life easier) and use all the indexing and other optimizations in your gist, while the other can be completely duck typed and only require that the input types support vector field operations (`+`, `* <scalar>` and `norm`), returning a `Vector{typeof(y0)}` instead of something more performant. If/when the performance bottlenecks of vectorized array code are fixed in base Julia so that they become equally performant, the vector specific method can be removed.

And by the way, I think a huge thank you is in place for putting time and effort into this re-write. Reading back in this thread it sounds like you're getting a lot of resistance that you don't deserve, just because a few of us happen to be picky - I hope you don't take it too hard :)

mauro3:
No worries and thanks!

I think the current version of the PR does what you describe, except that it avoids code duplication by wrapping any non-array thing `y0` as `[y0]` and calling the array solver.  Which then solves it as a single component vector system.  The operations performed on a single component are always just the scalar math operations, so it just works.  Does that make sense?

mauro3:
Ok, so I tidied everything up (but not quite ready).  The last commit of the PR is now how I think the interface/internals should be.  It works for array `y0`s and scalar `y0` as in this gist: https://gist.github.com/mauro3/d605bff8e11994a372d9

I'll write about what interface the scalars and arrays need to conform to later.  I realize this is the 10'000th commit but maybe someone still has some more time to look at this...

src/runge_kutta.jl
```diff
@@ -0,0 +1,509 @@
...
+    tsteps = length(tspan)
+    ys = similar(y0, Ey, dof, tsteps)
+    ys[:,1] = y0
+    tspan = convert(Vector{Et}, tspan)
+    # work arrays:
+    ks = similar(y0, Ef, dof, S)
+    ytmp = similar(y0, Ey, dof)
+    for i=1:length(tspan)-1
+        dt = tspan[i+1]-tspan[i]
+        ys[:,i+1] = ys[:,i] # this also allocates a bit...
```
pwl: Writing it explicitly as a for loop improves the performance for Euler by the factor of 2 and for rk4 by a factor of 1.2.  Not much but it's always something.

src/runge_kutta.jl
```diff
@@ -0,0 +1,509 @@
...
+        end
+    end
+    return tspan, ys
+end
+
+# calculates k[s]
+function calc_next_k!{N,S}(ks::Matrix, ytmp::Vector, y, s, fn, t, dt, dof, btab::TableauRKExplicit{N,S})
+    # Calculates the next ks and puts it into ks[:,s]
+    # - ks and ytmp are modified inside this function.
+    ytmp[:] = y
```
pwl: Another small improvement can be made by devectorizing this.

pwl: By the way, I noticed that the scalar version is much slower (5s vs 3s for fixed step rk4) than the vector version but I can't see why.

acroy: Could it be due to the additional anonymous function in line 200?

src/runge_kutta.jl
```diff
@@ -0,0 +1,509 @@
...
+    for ss=1:s-1, d=1:dof
+        ytmp[d] += dt * ks[d,ss] * btab.a[s,ss]
+    end
+    ks[:,s] = fn(t + btab.c[s]*dt, ytmp)
```
pwl: Adding a type assertion (I changed it to `ks[:,s] = fn(t + btab.c[s]*dt, ytmp)::typeof(ytmp)` but maybe someone knows a better way of doing this...) gives another improvement.  Although it requires a redefinition of `fn_` in line 200 to `fn_ = (t, y) -> [fn(t, y[1])]` for consistency.

````
# with type assertions
julia> @elapsed ODE.oderk_fixed((t,y)->y,1.0,linspace(0.0,1.0,10^7),ODE.bt_rk4)
6.424146062
julia> @elapsed ODE.oderk_fixed((t,y)->y,[1.0],linspace(0.0,1.0,10^7),ODE.bt_rk4)
2.396699645

# without type assertions
julia> @elapsed ODE.oderk_fixed((t,y)->y,1.0,linspace(0.0,1.0,10^7),ODE.bt_rk4)
6.690442561
julia> @elapsed ODE.oderk_fixed((t,y)->y,[1.0],linspace(0.0,1.0,10^7),ODE.bt_rk4)
3.603951279
````

(not much of a change for scalar version I guess it's because the additional allocation compensates the gain).

File: src/runge_kutta.jl [hunk 7]
```diff
@@ -0,0 +1,509 @@
...
+    #       the moment a FloatingPoint
+    # - Ey: eltype of y
+    # - Ef: eltype of fn(t,y)
+    #   --> both of these are set to typeof(y0[1]/tspan[1])
+    # - btab: tableau with entries converted to Et
+
+    Ey = typeof(y0[1]/(tspan[end]-tspan[1]))
+    Ef = Ey # this could be changed if function type become available
+
+    Et = promote_type(eltype(tspan), Float16)
```
pwl: I also don't get the promotion to Float16 (Float16 in particular?).  Maybe this should be an assertion instead?

src/runge_kutta.jl
```diff
@@ -0,0 +1,509 @@
...
+    f1     = similar(y0, Ey, dof)
+
+    # Option points determines where solution is returned:
+    if points==:specified
+        nsteps = length(tspan)
+        ys = similar(y0, Ey, dof, nsteps)
+        ys[:,1] = y
+        tspan_fixed = []
+    elseif points==:all
+        ys = similar(y0, Ey, 0)
```
acroy: Is it really necessary to handle the two cases differently? Can't you use the same approach for `points==:specified` and use `sizehint!` to preallocate the memory?

src/runge_kutta.jl
pwl: I guess so, I just didn't know that the cost of defining an intermediate anonymous function can have such impact.

src/runge_kutta.jl
mauro3: Thanks, good spot! Also fixing `fn_` is good.

src/runge_kutta.jl
mauro3: `ytmp[:] = y` is fast, at least for arrays (it uses copy! internally).  I tested this just today:
https://gist.github.com/mauro3/dfa1d24adaaa1bf48d7d

I'm aware that the scalar version is slower.  I haven't quite figured it out yet but I'll look into it.  One thing I noted that a lot of time was spent in `stepsize_hw92`, which actually needs some tweaking for scalars.

src/runge_kutta.jl
mauro3: Done.

src/runge_kutta.jl
mauro3: There is definitely room for improvement there.  The idea is to make it a floating point type.  This would avoid errors for `Vector{Int}` tspans.  And I thought that this would do the right thing: promote most things to `Float64` but keep the `Float32` and `Float16` as they are.  But it doesn't, so it's just plain wrong.  I'll come up with something more clever ;-)

mauro3: How about
```
    Et = eltype(tspan)
    @assert Et<:Real
    if !(Et<:FloatingPoint)
        Et = promote_type(Et, Float64)
    end
    if isleaftype(Et)
        warn("The eltype(tspan) is not a concrete type!  Change for better performance.")
    end
```
This keeps floating points as they are and promotes all others.  We can think about Rationals if someone files an issue.

src/runge_kutta.jl
mauro3: Yes, this is not quite right yet, but I'm not sure yet how to make it right.

src/runge_kutta.jl
pwl: You are correct, in your tests `ytmp[:]=y` seems to be even faster than the looped version.

src/runge_kutta.jl
pwl: Shouldn't it be `!isleaftype(Et)`?

src/runge_kutta.jl
mauro3: Actually `sizehint` will only allocate the memory needed for the pointers to the vectors, not the space needed for the vectors themselves.

acroy: Yes, but at the moment you will have at least one new copy per solution step anyways due to `F` not being mutating. So you don't need to preallocate the vectors.

mauro3: Yep, that's right. This is supported by the profiling I just did, which didn't show a difference.

acroy:
Your solvers return `Vector{typeof(y)}` and `Matrix{eltype(y)}` for "scalar" and array input, right? This breaks the current behavior, which isn't a big problem but we have to change the other solvers (e.g. `ode23s`) as well. I would suggest to either stick to the old behavior and change to Vector/Matrix in a separate PR for all solvers or to tag the current master as 0.2 and merge your PR afterwards. There are probably only a few user who will care, but we should try to keep the package consistent in some form.

mauro3:
Yes, I'm aware of this and still mulling over it.  But for sure it should be merged in an orderly fashion. I could prepare another PR to first do the API change.

What I would like to have is the following input -> output

1. scalar `y0` -> `Vector{typeof(y0)}` 
2. Vector input (e.g. `QuArray`) -> `Vector{QuArray}`
3. 2D vector input (e.g. `[1., 2.]'`) -> `Matrix{Float64}`

So, API of 1 & 2 are the same and are consistent with the current API.  3 is different and a slightly breaking change.  However the corresponding internals should be:

1. do everything as scalar.  Implement by wrapping `y0` as `[y0]` and calling (2)
2. do calculation element-wise in-place for each time step after initial creation of the current state vector
3. pre-allocate the output matrix.  then do all calculations element-wise in-place

But I'm not sure yet how to implement it such that 2&3 can share the algorithmic part of the code.

Apart from the technical issues, would this be a good API & internals?

acroy:
Sounds very reasonable to me.

tlycken:
I see a few possible problematic points here - that doesn't mean I think this is a bad idea, but I want us to both conceptually and actually in code before we merge changes like these. 

1. Implementing scalar solvers by wrapping in a vector isn't going to be type stable; for scalar y0 we want to return Vector{typeof(y0)} but 2) will return Vector{Vector{typeof(y0)}} if called with [y0] - I don't see a performant way of unwrapping that. (And wrapping scalars in a vector just to satisfy the api is going to be awful for performance anyway...) 

2. The performance gains in using a matrix for the output comes from the ability to preallocate it, right? How do we know how many steps to allocate space for?

3. There's a few things that were possible in the old api that won't work the same if we start treating any non-AbstractVector as a pseudo-scalar - for example, it becomes more difficult to use custom norms to calculate the residual in the adaptive methods, a feature that was explicitly asked for on the mailing list. 

I would much rather have the same api as before, with the optimizations in a specialization for the types for which they are applicable. I'd rather see some code duplication, since they won't actually be algorithmically equivalent anyway, than trying to fit everything into the optimized code with both shoehorn and vaseline...

acroy:
Maybe we could put the performance critical in-place updates in separate functions and keep the "algorithmic part" general? What we need is basically something like `add!(A,B,beta)` which does A + beta*B -> A and `copy!`. For the former we can provide a specialized version for vectors (I think there was/is some discussion to have a function like that in Base, but I can't find it). 

I tried this for `oderk_fixed` in [this gist](https://gist.github.com/acroy/29f6b4734b6ce66f66c1) (which is a bit hacky) and the performance seems to be comparable to Mauro's implementation. Maybe there is some caveat though... And of course I don't know how easy it is to do something similar for the adaptive solver.

mauro3:
Thanks for the feedback Tomas:

1. Yes, it needs unwrapping but it doesn't take long.  For example, a 10^7 long `Vector{Vector{String}}` takes 0.15s on my laptop.  I would expect any ODE solver would take at least 10x longer, probably much longer, to produce so many values.  Probably worse is the double anonymous function.
2. Fully preallocating would necessitate `points=:specified`, otherwise it will be `push!`.
3. The current API treats everything as a pseudo-scalar.  I think both `F` and `norm` should act on the whole state, i.e. pseudo-scalar, irrespective of whether in scalar, vector or matrix mode.  It is then up to the user to internally devectorize these functions if needed.  So, I think the step-size control can stay the same (having said this, the step control in this PR needs more work to get to that).

Thanks Alex for prototyping this!  The problem I see with your approach is that it only works with scalar `y0` which are `<:Number`, otherwise the overloading of `copy!` doesn't work.  Plus the overloading pollutes the `Base` functions with hacks.  This could be rectified by using our own local versions of `copy!`.  I just tried to update your version, pasted below your gist, and my experience was that it was quite confusing trying to do vectors and scalars at the same time...

So, my proposal to continue this PR, is to implement points 1 & 2 and leave 3 away for now.  Thus the API is completely unchanged.  I'm pretty sure that 3 can eventually be added onto 2 by using ArrayViews.jl.  The advantage of this, compared to Alex's prototype is reduced complexity, all it takes is:
```julia
function oderk_fixed(fn, y0, tspan, btab::TableauRKExplicit)
    # Non-arrays y0 treat as scalar
    fn_ = (t, y) -> [fn(t, y[1])]
    t,y = oderk_fixed(fn_, [y0], tspan, btab)
    return t, vcat_nosplat(y)
end
vcat_nosplat(y) =  eltype(y[1])[el[1] for el in y]
```
, compared to 8 function definitions in Alex's prototype. If we find that scalar performance is indeed too bad, then let's revisit Alex's approach.

src/runge_kutta.jl
mauro3: Added extra promotion as well.

src/runge_kutta.jl
```diff
@@ -0,0 +1,509 @@
...
+            for d=1:dof
+                ys[d,i+1] += dt * btab.b[s]*ks[d,s]
+            end
+        end
+    end
+    return tspan, ys
+end
+
+# calculates k[s]
+function calc_next_k!{N,S}(ks::Matrix, ytmp::Vector, y, s, fn, t, dt, dof, btab::TableauRKExplicit{N,S})
```
acroy: I guess it should be `ytmp::AbstractVector`?

acroy:
> The problem I see with your approach is that it only works with scalar y0 which are <:Number, otherwise the overloading of `copy!` doesn't work. Plus the overloading pollutes the Base functions with hacks.

Yeah, the situation with `copy!` and `similar` is not really optimal (although one might argue that `copy!` for Numbers and other bitstypes should be included in Base anyways). I think the nice aspect of my approach is, that the algorithm is separated from performance optimizations for specific types.

But I also think that we don't have to do this separation now; your proposal looks very reasonable to me. The biggest problem for pseudo-scalars is probably the additional anonymous function.

src/runge_kutta.jl
mauro3: The two cases are handled a bit better now.  I'll see whether I can improve it some more.

mauro3:
Just now I updated according to my last post.

src/runge_kutta.jl
mauro3: I think this is ok now.

mauro3:
This is it, point 1 & 2 done.  Probably needs some moving of things to other files, removal of old solvers.  And I want to add some more test, for instance Alex's example: https://gist.github.com/mauro3/d605bff8e11994a372d9

mauro3:
Life caught up, I'll work on this again in ~10 days.

berceanu:
What is still missing before we can merge this?

mauro3:
Sorry for having left this so long and thanks @berceanu for offering help.  I suspect that it is probably easiest if I finish this.  However, code review would certainly be good.  Also helpful would be if you could try out this branch on your problem and report back.

mauro3:
PR #70 adds a test for a scalar-like state.

mauro3:
I ran a performance test for the custom scalar type of #70 (with `endt=200*pi`)

```
| Fixed step | time |   MB |
|------------+------+------|
| ode4       |  0.1 |  7.6 |
| ode4_v2    | 0.16 | 13.6 |

| Adapt step | time |  MB |
|------------+------+-----|
| ode45_dp   | 0.08 |  94 |
| ode54_v2   |  0.5 | 211 |
|            |      |     |
| ode78_fb   | 0.07 |  84 |
| ode78_v2   |  0.4 | 148 |
```
The `_v2` are the new corresponding solvers.  So the trick of wrapping a scalar-like thing in `[]` and run that way is indeed pretty inefficient.  So, an approach as suggested by @acroy above might be better.

acroy:
I am kind of surprised that the difference is that large ... On the other hand for array-like input this PR should be (much) faster, right? So the question is if we want everything now or if we are willing to accept that the less common case (?) is slower and address that in a separate PR?

acroy:
@mauro3 : Could you provide some benchmarks for the number and array input cases?

I would suggest the following:
- We tag the current master (or rather after merging #70) as v0.2. If for some reason anyone needs the current behavior it is at least easy to get it.
- We merge this PR (when Travis is happy) and accept the performance regression for custom types. For most people this will probably bring an improvement though (given the benchmarks actually support this).
- We work on the alternative approach to unify number/array/custom type input in a separate PR. This might need some discussion/synchronization with Julia core anyways.

Does that make sense?

ViralBShah:
From the sidelines, I see a lot of work has happened. I feel it is best to tag a release and note some of these things in the README.

pwl:
+1 for @mauro3.  I also agree with @ViralBShah that some comment in the README would be nice.

mauro3:
The performance is better with the new solvers on vector-problems, with the latest version the performance is slightly better than what I posted above: https://github.com/JuliaLang/ODE.jl/pull/68#issuecomment-94713957

For a more simple test case (the last of the runtest.jl tests, https://gist.github.com/mauro3/1a4d2eb85238dbfb9251) I get:
```
| Fixed step | time |  MB |
|------------+------+-----|
| ode4       | 0.15 | 126 |
| ode4_v2    | 0.06 |  38 |

| Adapt step |  time | MB |
|------------+-------+----|
| ode45_dp   |  0.09 | 71 |
| ode54_v2   |  0.04 | 14 |
|            |       |    |
| ode78_fb   | 0.037 | 37 |
| ode78_v2   | 0.025 |  4 |
```
So, in summary about 2x faster and uses a lot less memory.

I can add something to the readme.

mauro3:
Before merging this, two items are left:
- removing old solver
- squashing commits

Let me know if you think this is mergable and I'll finish up those two points.

Also, should I keep the RK solvers in their own file or include them in the single `ODE.jl` file?

mauro3:
The travis failure on 0.4 is really strange and I cannot reproduce it locally, neither with a fresh build nor with the nightlies.  In the latest comit I added printout for the numerical error.  0.3 produces:
```
using ode23_bs
0.0001247361961962401
using ode4
0.00418389695346022
using ode45_dp
1.4920637694704197e-5
using ode45_fb
2.799481393722658e-5
using ode45_ck
0.00013077221580326537
using ode4_v2
1.0169642905566434e-13
using ode21_v2
1.90809077746934e-5
using ode45_v2
8.304742954790001e-5
...
```
whereas 0.4
```
using ode23_bs
0.0001247361961962401
using ode4
0.00418389695346022
using ode45_dp
1.4920637694704197e-5
using ode45_fb
2.799481393722658e-5
using ode45_ck
0.00013077221580326537
using ode4_v2
1.0169642905566434e-13
using ode21_v2
0.2659455163733535
# test-error
```
So, all the numerical errors are exactly identical except for `ode21_v2`.  One thing I omitted is the simd warnings occuring on 0.4. (see https://travis-ci.org/JuliaLang/ODE.jl/jobs/63471018) which do not happen on my local machine.  Any ideas?

src/runge_kutta.jl
```diff
@@ -0,0 +1,569 @@
...
+## Tableaus for explicit methods
+# Fixed step:
+const bt_feuler = TableauRKExplicit(:feuler,(1,), Rational{Int64},
+                                    zeros(Int,1,1),
+                                    [1]',
+                                    [0]
+                                    )
+const bt_midpoint = TableauRKExplicit(:midpoint,(2,), Rational{Int64},
+                                      [0  0
+                                       .5  0],
```
acroy: `1//2` for consistency? (Same two lines below)

acroy:
On my somewhat oldish 0.4 (26 days old) this also works locally for me. So I don't really know. Do the other adaptive solvers work if you remove `ode21_v2` from the tests?

acroy:
Regarding the file structure, I would prefer to keep the RK solvers in a separate file. We have to decide where `hinit` should go though, since it is also used by `ode23s`. And actually I would move the Butcher tables into yet another file.

mauro3:
Travis worked fine on the other tests: https://travis-ci.org/JuliaLang/ODE.jl/builds/63480677 Most peculiar.

I will now update the file structure as suggested by @acroy and rebase.

mauro3:
Removed old solvers and squashed commits.  Ready to merge as far as I'm concerned.
(Testing my luck, I added the test-case which failed on 0.4 before.  I'll update if Travis fails.)

mauro3:
One last comment on performance.  I now support the `norm` keyword in the solver as well. As this is adds another higher order function call it impacts performance.  For instance a high precision run with `ode54` without `norm` keyword:
```
Running test 7 of 7: sig. digits= 6.978087256702925, walltime= 0.007011848s, memory= 4.0MB
```
with `norm` keyword:
```
Running test 7 of 7: sig. digits= 6.978087256702925, walltime= 0.009105477s, memory= 6.0MB
```
I suggest to do this differently using dispatch.  But that is for another PR.

src/runge_kutta.jl
```diff
@@ -0,0 +1,473 @@
+# Explicit Runge-Kutta solvers
+##############################
+# (Hairer & Wanner 1992 p.134, p.165-169)
+
+###########################################
+# Tableaus for explicit Runge-Kutta methods
+###########################################
+
+immutable TableauRKExplicit{Name, S, T} <: Tableau{Name, S, T}
```
berceanu: I get `ERROR: type: instantiate_type: expected TypeConstructor, got Type{(Any...,)}` on this line with Julia 0.3.8-pre+2.

mauro3:
@berceanu, I just pushed a few minor fixes but probably not to your problem.  Is that error happening when running the tests?  Could you post the full back-trace?

acroy:
For me everything works locally (I am also on 0.3.8-pre+2).

acroy:
Travis failure for 0.4 seems unrelated.

mauro3:
It doesn't even look like there was an error on travis. What went wrong?

acroy:
Maybe a Travis hiccup ... I restarted the 0.4 run ...

berceanu:
Here is the full trace:

```
ERROR: type: instantiate_type: expected TypeConstructor, got Type{(Any...,)}
  in include at ./boot.jl:245
  in include_from_node1 at ./loading.jl:128
  in include at ./boot.jl:245
  in include_from_node1 at ./loading.jl:128
  in reload_path at loading.jl:152
  in _require at loading.jl:67
  in require at loading.jl:51
  in include at ./boot.jl:245
  in include_from_node1 at ./loading.jl:128
while loading /home/berceanu/.julia/v0.3/ODE/src/runge_kutta.jl, in 
expression starting on line 9
while loading /home/berceanu/.julia/v0.3/ODE/src/ODE.jl, in expression 
starting on line 148
while loading /home/berceanu/Development/topo-condensate/runge-kutta.jl, 
in expression starting on line 1
```

berceanu:
btw, the line in my code
```
runge-kutta.jl, in expression starting on line 1
```
is `using ODE`

acroy:
@berceanu : That is kind of strange. Can you try to run just the tests of this branch in a new session?

mauro3:
@berceanu thanks for trying this out! So you get the error by just `using ODE`?

Are you on the latest commit of this branch (on c8ad5f712df2e0b6)? I squashed and force pushed, so a simple pull on your side will not work.  Easiest would be to to delete the ODE directory and clone it again.  Or something like https://stackoverflow.com/questions/1125968/force-git-to-overwrite-local-files-on-pull should work.

mauro3:
I removed `ode54` and renamed it to `ode45`.  So the exported API should be non-breaking.

@berceanu: I definitely cannot reproduce your error.  I ran your topo-condensate code and it works for me (well the `@test_approx_eq_eps` tests fail).  Let me know if you have any luck with a fresh clone of ODE.jl.

berceanu:
I did

```
rm -rf .julia/v0.3/ODE/
git clone git@github.com:mauro3/ODE.jl.git ODE
git checkout m3/rk

```

and using ODE still gives the same error. Am I missing something?

mauro3:
I did those steps as well, just to be sure, and still no error here.  I also tried with a build of the realease-0.3 branch with:
```
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "help()" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.3.9-pre+27 (2015-05-20 15:14 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit 8418a3a* (1 day old release-0.3)
|__/                   |  x86_64-unknown-linux-gnu
```
with no error.


What operating system are you on?  You compiled from source with the `release-0.3` branch, right? Can you try it with 0.3.8 or the current head of relase-0.3?  Or from a julia-0.3 binary?

If you got a 0.4 version around, can you try that?

berceanu:
OK so I did a `Pkg.update()`, which bumped `Compat` from 0.3.7 to 0.4.4 
(among other things) and now `using ODE` no longer crashes with the code 
in your branch.

mauro3:
Cool, thanks!  I think that must have been the new Tuple types which were not included in the old compat.  Are those assertion errors I got running your code anything to worry about?

berceanu:
Well, it looks like with the new ode45, for my test case I get the same 
results for aout[end] with accuracy of 1e-6. The test that fails 
required 1e-15, so I guess it's not something to worry about.
The other difference seems to be the number of time steps taken, before 
I had length(tout) = 11 and with the new solver I have 12. Again, 
probably nothing to fret over.

kmsquire:
> I think that must have been the new Tuple types which were not included in the old compat.

Perhaps add a minimum version to REQUIRES for Compat.jl?

mauro3:
@berceanu how does the performance compare?  Time and memory allocated?

@kmsquire: done.  Also increased julia version to 0.3

acroy:
Ok. I just tagged a new version for ODE.jl/master. From my side we can merge this. @pwl also gave +1 and there were no objections. Let's do it then.

acroy:
Merged by popular demand :-)

mauro3:
:-)

pwl:
Great!  Tanks @mauro3!

acroy:
@mauro3 : I knew I forgot something: could you please update LICENSE.md?

mauro3:
It's here #73. Let me know what you think.

tlycken:
:100: :+1: :exclamation: 

Great job landing this!

PR #70, opened 2015-05-20
Added @acroy's test for custom scalar-like state

mauro3:
Also modified ode4 to use normal arithmetic, not `.*`-like

### DIFF
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -279,12 +279,12 @@ function ode4(F, x0, tspan)
     for i = 1:length(tspan)-1
         # Compute midstep derivatives
         midxdot[1] = F(tspan[i],         x[i])
-        midxdot[2] = 2*F(tspan[i]+h[i]./2, x[i] + midxdot[1].*h[i]./2)
-        midxdot[3] = 2*F(tspan[i]+h[i]./2, x[i] + midxdot[2].*h[i]./2)
-        midxdot[4] = F(tspan[i]+h[i],    x[i] + midxdot[3].*h[i])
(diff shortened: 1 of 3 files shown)


mauro3:
Also, unrelated to the main part of the PR changed `copy` to `deepcopy` in ode23s.  I think that is more correct but I can take it out.

acroy:
Thanks! This looks good. I am not 100% sure about `deepcopy` vs `copy`, but you are probably right. Do I get this right, that this test case doesn't work with `ode23s`? In that case we should have a new issue for fixing it.

mauro3:
Yes, ode23s and the fixed step Rosenbock solvers use indexing and other bits. 

If you got a vector of mutables then a shallow copy will leave them the same, which might lead to problems?

Also, this is also part of #68.  If we merge that PR soon then we could close this and get it via 68.

acroy:
> Yes, ode23s and the fixed step Rosenbock solvers use indexing and other bits.

The way the Jacobian is calculated prevents this (there is actually a "FIXME" comment for this - probably written by me :-).

> If you got a vector of mutables then a shallow copy will leave them the same, which might lead to problems?

Yes, it's better to use `deepcopy` then.

So from my side this can be merged.

ViralBShah:
You guys know best when to merge.

mauro3:
If PR #68 gets merged soon then I'd prefer to ditch this to save me some conflicts in rebasing #68.  Otherwise merge.

mauro3:
Closing as the test was merged with #68

PR #72, opened 2015-05-27
WIP: Rosenbrock(-W) methods

mauro3:
This is work in progress of me and @jedbrown to implement Rosenbrock(-W) methods.  It uses a different interface to what ODE.jl has: implicit in-place functions.  Works for ODE and DAEs.  At the moment two methods are implemented: [Rodas3](http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/TS/TSROSWRODAS3.html#TSROSWRODAS3) and [RA34PW2](http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/TS/TSROSWRA34PW2.html#TSROSWRA34PW2).

This will take some time until it is ready to be considered for a merge.  But it works now (see `examples/` folder) and might be of use to someone. The simple benchmarks in the example folder suggests that these solvers can be faster and more memory efficient than DASSL and `ode23s`, e.g. van der Pol:

```
ode23s with jacobian:
elapsed time: 0.025410477 seconds (25_194_952 bytes allocated)
ode_rosw with jacobian:
elapsed time: 0.00535883 seconds  (2_462_312 bytes allocated)
ode_rosw with numeric jacobian:
elapsed time: 0.013636325 seconds (9_859_344 bytes allocated)
```

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.3
 Polynomials
 Compat 0.4.1
+MatrixColorings
--- a/examples/bruss1d.jl
+++ b/examples/bruss1d.jl
@@ -169,34 +169,37 @@ ic[1:2:2N] = u0(x)
 ic[2:2:2N] = v0(x)
 tspan = T[0.0, 10.0] # integration interval
 tspan0 = T[0.0, 0.001] # integration interval
+tspan1 = T[0.0, 0.3] # integration interval
 
 #t,ydassl = dasslSolve(fn, ic, tspan, jacobian=jac)
 
 ## ode23s
-# tout, yout = ode23s(fn_ex, ic, tspan0, jacobian=jac_ex)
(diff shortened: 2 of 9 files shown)


mauro3:
Large sparse Jacobians can now be numerically approximated using coloring with https://github.com/mauro3/MatrixColorings.jl. The bruss1d.jl example shows it in action.

pwl:
Wow, the performance boost is really impressive.  What is the increase in performance for smaller systems?  Do you think it could be used in DASSL?

mauro3:
Thanks :-)

The coloring only really helps for sparse Jacobians (try setting N=5000 in the bruss1d example and use the numeric Jacobian in DASSL). I suspect small sparse systems may still be slower.  You should be able to pretty much just plug it into DASSL except that I used in-place functions, see: https://github.com/mauro3/ODE.jl/blob/m3/rosw/src/rosw.jl#L465

PR #74, opened 2015-06-10
Fixed bug in runge-kutta

mauro3:
When intital time step overstepped `tspan[end` then it didn't work correctly in the RK solvers. @acroy: can you check that the modification to hinit is ok for ode23s?


acroy:
Thanks for fixing this. I think it should be Ok for `ode23s`. There is already a check if the step would go beyond `tspan[end]` and the step-size is adjusted to `tspan[end]-tspan[1]` if necessary.

mauro3:
I had that check in place as well, except for the first step.

acroy:
In `ode23s` the check is at the beginning of the loop and thus also applies to the first step.

IMO the PR looks good.

PR #76, opened 2015-09-17
Fix 0.4 deprecated bindings

yuyichao:
Also removed some white space and explicitly import operators from Base.


tlycken:
LGTM :+1:

ViralBShah:
Should probably tag as well, right?

acroy:
Yes. This should be tagged as v0.2.1.

PR #79, opened 2015-10-14
While loop in ode23s correctly handles negative initial time

ryanpdwyer:
Currently, running `ode23s` when `abs(tspan[1]) > abs(tspan[end])` will prevent the main while loop from running. For example, 

```julia
ODE.ode23s((t,y)->[y[2]; -y[1]], [1., 0.], [-5., 0])
```

just returns the initial time point on master. This commit adds a test and changes the while loop condition to work for any `tspan`.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -60,6 +60,10 @@ for solver in solvers
     @test maximum(abs(ys-[cos(t)-2*sin(t) 2*cos(t)+sin(t)])) < tol
 end
 
+# Test negative starting times ODE.ode23s
+@assert length(ODE.ode23s((t,y)->[-y[2]; y[1]], [1., 2.], [-5., 0])[1]) > 1
+
+
 # rober testcase from http://www.unige.ch/~hairer/testset/testset.html
 let
     println("ROBER test case")
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -250,8 +250,8 @@ function ode23s(F, y0, tspan; reltol = 1.0e-5, abstol = 1.0e-8,
 
     # initialization
     t = tspan[1]
-    tfinal = tspan[end]
 
+    tfinal = tspan[end]
 
     h = initstep
     if h == 0.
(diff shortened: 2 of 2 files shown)


acroy:
Well spotted! Thanks. I think it should be sufficient, to change
```
while abs(t) < abs(tfinal) && minstep < abs(h)
```
to
```
while abs(t - tfinal) > 0 && minstep < abs(h)
```
Note that the direction of integration is already determined and stored in `tdir`. But maybe I am missing something here ...

ryanpdwyer:
That results in a test failure,

```
ERROR: LoadError: DomainError:
Exponentiation yielding a complex result requires a complex argument.
Replace x^y with (x+0im)^y, Complex(x)^y, or similar.
 in ^ at math.jl:275
 in ode23s at /Users/ryandwyer/.julia/v0.4/ODE/src/ODE.jl:336
 [inlined code] from /Users/ryandwyer/.julia/v0.4/ODE/test/runtests.jl:50
```

The line where the error occurs is,

```julia
h = tdir*min( maxstep, abs(h)*0.8*(delta/err)^(1/3) )
````

during the test,

```
t,y=solver((t,y)->y, 1., [1:-.001:0;])
```

**Comment by acroy** [comment]
Yes, in line 302 it currently reads
```
err = (h/6)*norm(k1 - 2*k2 + k3)
```
which should anyways be
```
err = (abs(h)/6)*norm(k1 - 2*k2 + k3)
```
(if `h` is negative `(delta/err)^1/3)` becomes complex).

**Comment by ryanpdwyer** [comment]
I added `abs(h)` in that line, and before the loop,

```julia
h = tdir * min(abs(h), maxstep)
```

acroy:
Looks good. Thanks again!

test/runtests.jl
```diff
@@ -60,6 +60,10 @@ for solver in solvers
     @test maximum(abs(ys-[cos(t)-2*sin(t) 2*cos(t)+sin(t)])) < tol
 end
 
+# Test negative starting times ODE.ode23s
+@assert length(ODE.ode23s((t,y)->[-y[2]; y[1]], [1., 2.], [-5., 0])[1]) > 1
+
+
```
mauro3: Sorry, too late to the party, but should/could this test not be run for all of the solvers?

PR #84, opened 2015-12-31
Remove extra whitespace and indent properly

pranitbauva1997:


ViralBShah:
The return is not indented correctly now. Let's not do simply indenting changes. This file already has 2 different indenting whitespace settings, and that's ok for now.

PR #88, opened 2016-02-20
Enable precompilation

jgoldfar:


ViralBShah:
@tkelman Is this still the preferred way to do the precompliation?

tkelman:
Either this or `isdefined(Base, :__precompile__)`

Using `isdefined` has the benefit that it would work on `NO_GIT` builds of older 0.4-dev where version numbers are not accurate, but that's probably a rare case.

acroy:
Certainly nice to have! I don't have a strong preference, which version we should use, but Tony's looks more elegant ...

jgoldfar:
Good to know; I'll update the PR. I would note that Compat.jl does it this way, so it may be worth updating that as well

tlycken:
`isdefined` has the added benefit of being *forwards* compatible, in case the mechanism for precompilation is changed in a future version of Julia (at least in the sense that nothing will throw errors; precompilation will silently break, but that won't make the package unusable).

jgoldfar:
Test for when to enable precompilation is now updated.

PR #89, opened 2016-02-20
also test 0.3 on travis

tkelman:
still supported according to REQUIRE, so should still be tested


acroy:
Thanks! I would say we should keep support of 0.3 until we need some of the newer features, so this change certainly makes sense.

PR #92, opened 2016-03-12
Update Simple_Differential_Equation.ipynb

jpfairbanks:
Avoid using `start` as a variable in example

### DIFF
--- a/examples/Simple_Differential_Equation.ipynb
+++ b/examples/Simple_Differential_Equation.ipynb
@@ -92,7 +92,7 @@
    "outputs": [],
    "source": [
     "# Initial condtions -- x(0) and x'(0)\n",
-    "const start = [0.0; 0.1]\n",
+    "const starty = [0.0; 0.1]\n",
     "\n",
     "# Time vector going from 0 to 2*PI in 0.01 increments\n",
     "time = 0:0.1:4*pi;"
@@ -113,7 +113,7 @@
    },
    "outputs": [],
    "source": [
-    "t, y = ode45(f, start, time);"
+    "t, y = ode45(f, starty, time);"
    ]
   },
   {


jpfairbanks:
I went through this example and got this error if we avoid the variable `start` it shouldn't happen.
````julia
julia> ode45(f, start, trange)
ERROR: MethodError: `/` has no method matching /(::Function, ::Float64)
Closest candidates are:
  /(::Float64, ::Float64)
  /(::Complex{T<:Real}, ::Real)
  /(::BigFloat, ::Union{Float16,Float32,Float64})
  ...
 in make_consistent_types at /home/james/.julia/v0.4/ODE/src/ODE.jl:124
 in oderk_fixed at /home/james/.julia/v0.4/ODE/src/runge_kutta.jl:180
 in oderk_fixed at /home/james/.julia/v0.4/ODE/src/runge_kutta.jl:168
 in ode45 at /home/james/.julia/v0.4/ODE/src/runge_kutta.jl:163
```

acroy:
Using `start` as a variable shouldn't cause problems. Did you try to restart julia?

In any case it is good practice to post a complete example (possibly including the versions of Julia and ODE), which shows the error. If I do the following in Julia v0.4 and ODE v0.2.1:
```
using ODE

function f(t, y)
           # Extract the components of the y vector
           (x, v) = y
           
           # Our system of differential equations
           x_prime = v
           v_prime = -x
           
           # Return the derivatives as a vector
           [x_prime; v_prime]
end

const start = [0.0; 0.1]
time = 0:0.1:4*pi

t, y = ode45(f, start, time);
```
There is no problem:
```
julia> y[end]
2-element Array{Float64,1}:
 -0.00663187
  0.0997793 
julia> t[end]
12.5
```

pjpmarques:
Hi. Any conclusion on this? `start` shouldn't be the cause of the problem.

ChrisRackauckas:
`start` may not be the cause of the problem due to ODE, but having a variable named `start` can easily cause problems in other cases. @jpfairbanks probably had a problem due to something defining an [array interface](http://docs.julialang.org/en/release-0.4/manual/interfaces/) (which is why the error is due to a function). 

Because these are so common, it's a good idea to not have `start` as a variable in an example since it has a common other use and is thus begging for trouble. Changing the name to `starty` is harmless but gets rid of any possibility of this overlap. I'll wait for a second opinion but vote to merge.

@pwl @acroy @mauro3 ?

ChrisRackauckas:
Given the +1's and that it's pretty harmless, I'm merging.

PR #93, opened 2016-03-31
Component-wise relative and absolute tolerances

sonVishal:
Hi,
I have tried to work out the component-wise relative and absolute tolerances for the non-stiff solvers based on 'Solving Ordinary Differential Equations I by Hairer and Wanner' and component-wise absolute tolerance combined with strictly scalar relative tolerance for the stiff solvers based on the MATLAB ode23s solver.
Since the interface checks does not handle vectors yet, the interface checks fail.
I have also changed the value of the error for the Robertson test since even Matlab's ode23s cannot produce a result with error strictly less than 2e-10. I got 2.0423543e-10 using the same options as used over here. Hence the change.
Please review.

Thanks and regards,
Vishal Sontakke,
Technical University of Munich.


src/ODE.jl
```diff
@@ -66,10 +66,10 @@ function hinit(F, x0, t0, tend, p, reltol, abstol)
     # Returns first step, direction of integration and F evaluated at t0
     tdir = sign(tend-t0)
     tdir==0 && error("Zero time span")
-    tau = max(reltol*norm(x0, Inf), abstol)
-    d0 = norm(x0, Inf)/tau
+    tau = max(reltol.*abs(x0), abstol)
+    d0 = norm(x0,./tau Inf)
```
acroy: This looks wrong. Should probably be `x0./tau, Inf`.

sonVishal: Ah yes. I had fixed that in my code. Didn't update. Sorry. Will commit it again.

acroy:
Thanks for the PR! You have accidentally deleted the .travis.yml file, which prevents us from getting Travis feedback. Please put it back.

sonVishal:
I added the file and also fixed the error as you pointed out.

src/ODE.jl
```diff
@@ -251,6 +251,17 @@ function ode23s(F, y0, tspan; reltol = 1.0e-5, abstol = 1.0e-8,
 
     # initialization
     t = tspan[1]
+    
+    # Component-wise Tolerance check similar to MATLAB ode23s
+    # Support for component-wise absolute tolerance only.
+    # Relative tolerance should be a scalar.
+    @assert length(reltol) == 1 "Relative tolerance must be a scalar"
```
acroy: Why don't you use `isa(reltol, Number)` to check if it is a scalar? This also concerns the other asserts.

sonVishal: I wasn't aware of this function. I just wanted the code to stop if we do not give the proper input.
You mean instead of checking the lengths I simply use isa() function for the asserts?

acroy: Ok. I just realized that `length` returns `1` for a scalar.

tlycken: This would be much better handled with a type spec on the function parameter, ie `reltol::Number`.

src/ODE.jl
```diff
@@ -66,10 +66,10 @@ function hinit(F, x0, t0, tend, p, reltol, abstol)
     # Returns first step, direction of integration and F evaluated at t0
     tdir = sign(tend-t0)
     tdir==0 && error("Zero time span")
-    tau = max(reltol*norm(x0, Inf), abstol)
-    d0 = norm(x0, Inf)/tau
+    tau = max(reltol.*abs(x0), abstol)
```
tlycken: Rather than using abs here, we should allow using a custom norm function. Look at how eg ode23s does this, and pass that norm along to hinit as well.

src/ODE.jl
```diff
@@ -300,11 +311,11 @@ function ode23s(F, y0, tspan; reltol = 1.0e-5, abstol = 1.0e-8,
         F2 = F(t + h, ynew)
         k3 = W\(F2 - e32*(k2 - F1) - 2*(k1 - F0) + T )
 
-        err = (abs(h)/6)*norm(k1 - 2*k2 + k3) # error estimate
-        delta = max(reltol*max(norm(y),norm(ynew)), abstol) # allowable error
+        threshold = abstol/reltol # error threshold
+        err = (abs(h)/6)*norm((k1 - 2*k2 + k3)./max(max(abs(y),abs(ynew)),threshold),Inf) # error estimate
```
tlycken: err has suddenly changed meaning here - it is now what it was before divided by something else. Should it have another name, too?

test/runtests.jl
```diff
@@ -65,6 +65,10 @@ end
 
 
 # rober testcase from http://www.unige.ch/~hairer/testset/testset.html
+# Changing the test value.
+# Even the Matlab ode23s does not give the error to be less than 2e-10.
+# The value comes out to be approximately 2.042354e-10.
+# Hence changing the test value to 2.1e-10
```
tlycken: Did this test pass before? If so, why doesn't it now? I'm sure it might be justified to increase the tolerance a little, but then I'd like to see some better motivation than just "Matlab doesn't do better".

src/ODE.jl
sonVishal: This would only apply to reltol in the case of ode23s. For the other tolerances it would not be possible since they can be either a scalar or a vector of the same length as the problem.

src/ODE.jl
sonVishal: This function is not specific to ode23s. This function has been written as per 'Solving Ordinary Differential Equations I' by Hairer and Wanner and they simply use the absolute values of each component. The 'tau' used in this code is simply the scaling factor 'sc' used in the book.

src/ODE.jl
sonVishal: In the MATLAB ode23s they use 'err' as the name. It is not exactly the error but something similar to a scaled error. Hence I kept the same name. I could change it if it is really required.

test/runtests.jl
sonVishal: The test passed before I made changes.
I believe that the test fails because different step sizes are now chosen by the solver as compared to the previous code. I think we need to find better step size control.

The test did pass when I reduced the tolerances to 1e-9 from the original value of 1e-8.

src/ODE.jl
sonVishal: Can I ask why abs(x0) is not suitable over here?

src/ODE.jl
tlycken: Sure. But in the specific case handled by this line, since length 1 is hard-coded anyway, I think it would be much clearer.

sonVishal: Alright. I will make this change.

src/ODE.jl
tlycken: Since we're letting the user pass a custom norm function into the solver, we should use that here too - it's an oversight that this isn't already the case. I just think that if we touch these lines anyway, we might as well fix that (and if we don't, we should keep using norm here to make the transition easier later).

src/ODE.jl
tlycken: It's not required - I just didn't follow the details of what's going on with these few lines, and wondered if naming was contributing to my confusion :)

src/ODE.jl
sonVishal: Umm, this has nothing to do with the norm. As per the comments hinit is based on 'Solving Ordinary Differential Equations I' by Hairer and Wanner page 169 - The tau in the code is the same as the 'sc' variable in the book which is component-wise rather than a norm.

test/runtests.jl
tlycken: Nice, I think this change is safe to make, though. The comment could be left out though, as it's more a comment on the change than on the code itself.

src/ODE.jl
sonVishal: Ah. I should probably make a comment saying that it is a scaled error. Will add this comment to avoid any further confusion.

test/runtests.jl
sonVishal: Alright. I will remove the comments as well. :+1:

acroy:
Is there a fundamental reason why `ode23s` cannot have component-wise relative errors? I think one of the goals of this package is providing a uniform interface for different solvers (if possible).

Could we use a function for calculating the error estimate and facilitate mult. dispatch to distinguish between scalar and vector tolerances? That way we could keep the the old behavior for the scalar-like input and also avoid creating a tolerance vector for scalar `abstol`.

sonVishal:
I couldn't find a fundamental reason behind such a restriction.

In theory `reltol` can also be specified component-wise.

I was able to implement component-wise `reltol` for `ode23s` similar to the non-stiff solvers.

I have tried to restore the old behaviour when `reltol` and `abstol` are both scalars.
Let me know if another function is really necessary for this.

acroy:
Provided type inference works (`reltol` and `abstol` are keywords), separate functions and multiple dispatch allow you to avoid the `if` conditionals, which might have an impact on performance.

sonVishal:
Ah yes. I will try to update one with a function to get the error using multiple dispatch. :+1:

src/ODE.jl
stevengj: If you use an L-infinity norm, doesn't that correspond to component-wise error tolerances?

sonVishal: L-infinity would simply take the maximum over all the components.

Sometimes the solution is more sensitive w.r.t. only a component and you would thus want to control that component with a stricter tolerance and have relaxed tolerances for the other tolerances.

ChrisRackauckas:
What's the current status on this?

pwl:
We haven't planned for this in #49, but we should keep this in mind as a long term goal (after we merge #49).  I think we should close this for now and open an issue instead.

src/ODE.jl
ChrisRackauckas: The "Hairer norm" used here isn't necessarily a norm because of how it it does the normalization.  But the tau here is not the standard sc since sc is the sum of the adjusted tolerances, not the max. In practice this shouldn't matter much though.

ChrisRackauckas:
I don't think this should be closed due to an upcoming PR which doesn't include the functionality. It looks to me like this is complete, though I don't see a test. If @sonVishal adds some tests to show that it's working properly I'd vote to merge. How it relates with #49 is a separate issue.

sonVishal:
I will write a test case. I tested it myself but will include a test. Give me a week or so since I am currently very busy working on 2 different projects.

sonVishal:
I simply changed the relative and absolute tolerances for the ROBER test case to arrays and the test passed on my system.

Let me know if there is a need for another test case.

PR #96, opened 2016-05-25
Added test for using eltype(tspan)==Int and typeof(y0)=Int

mauro3:
We previously had problems with calls of the sort `ode((t,y)->2t, 0., [0,1])` where `tspan` is a `Vector{Int}`.  This adds a test to check that the solvers support this call.  It showed that `ode23s` didn't and also fixes this.

### DIFF
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -120,8 +120,8 @@ function make_consistent_types(fn, y0, tspan, btab::Tableau)
     # On container: eltype, promote_type
     # On time container: eltype
 
-    Ty = typeof(y0)
     Eyf = typeof(y0[1]/(tspan[end]-tspan[1]))
+    Ty = typeof(similar(y0, Eyf, 1))
 
     Et = eltype(tspan)
     @assert Et<:Real
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -265,7 +265,8 @@ function ode23s(F, y0, tspan; reltol = 1.0e-5, abstol = 1.0e-8,
     h = tdir * min(abs(h), maxstep)
 
     y = y0
-    tout = Array(typeof(t), 1)
+    # below promotion should maybe be replaced by using make_consistent_types
+    tout = Array(promote_type(typeof(t), Float64), 1)
     tout[1] = t         # first output time
     yout = Array(typeof(y0), 1)
     yout[1] = deepcopy(y)         # first output solution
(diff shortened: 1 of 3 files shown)


mauro3:
Added test and fix for using an integer as `y0`.  See:
https://groups.google.com/d/msg/julia-users/Mn0OrZLGNkg/cCQE7UH3BwAJ

pwl:
Is `Vector{Int}` the only ugly type that we want to support?  What about `Vector{BigInt}`, or a crazy mix like `Number[1,1.1,BigInt(1)]`?  How far do we go with this functionality?  Also I kind of agree that it would be nice to accept `Vector{Int}` along with other variations but the approach with `make_consistent_types` is never going to be type stable.  Is there no other way to implement this?  How about something along the lines of
```
ode{T<:AbstractFloat}(t0::AbstractVector{T})=print(t0)   # the actual implementation
ode(t0::AbstractVector)=f(float(t0))
```
so that instead of supporting `t0::AbstractVector{Number}` we would restrict the implementation to `t0::AbstractVector{AbstractFloat}`.  I don't know if this makes any sense but I like it slightly better then `make_consistent_types` approach.

mauro3:
I don't think it needs to be type stable if we use barrier-functions: if the type on which the iterator is based on is constructed using leaftypes only then it should be fine and it doesn't matter that figuring out the leaftypes is done by an type-unstable function.  We may want to draw the line at things like `Number[1,1.1,BigInt(1)]`.

I think one of the points of having pure Julia implementations of ode-solvers is to be able to use them with any Julia type (as long as that type supports certain operations).

ChrisRackauckas:
As mauro3 noted, you don't need type-stability if you shield the main loop inside another function which is type-stable. However, since that's not currently done, there is a chance that this will cause everything to "Box" and maybe cause issues with type-inference (maybe, because sometimes the boxing isn't an issue). Did you benchmark and check the native code for Int and Float64 inputs at all?

However, I don't think hard-coding Float64 is ever a good idea. For example, I'm pretty sure this will throw an `InexactError()` if someone is trying to use BigFloats, ArbFloats, etc. That's not a problem that should addressed here since this is making `ode23s` consistent with the other functions, but it's something to think about as a future improvement.

pwl:
We decided not to go this way and we are now throwing an error if there are any conversions to be made by the solver: https://github.com/JuliaODE/ODE.jl/blob/dev/test/top-interface.jl#L78-L84.  Any conversions that we come up with were simply too ambiguous and arbitrary, it's better if the user does them explicitly.

It's still a little early to do any benchmarks, but I have them in mind and we will do the optimization once we are happy with the current implementation of #49.

ChrisRackauckas:
Should this be closed then?

pwl:
I would say so.

PR #97, opened 2016-05-30
On tableau was in terms of 1/2 instead of 1//2, etc.

mauro3:
Used `/` instead of `//`.  This showed up when running with `BigFloat`.


ChrisRackauckas:
This is uncontroversial "bugfix". Tests pass and everything looks good. Merging.

PR #98, opened 2016-05-30
WIP: running tests over Float32 and BigFloat too

mauro3:
This is basically done, except that https://github.com/JuliaLang/julia/issues/16667 is blocking this.  That issue means that a few `deepcopy` had to me turned into `copy`.  Probably should wait until that is resolved and make them `deepcopy` again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ julia:
     - 0.4
     - release
     - nightly
+matrix:
+  allow_failures:
+    - julia: nightly
 git:
     depth: 999999
 notifications:
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -62,7 +62,7 @@ Base.convert{Tnew<:Real}(::Type{Tnew}, tab::Tableau) = error("Define convert met
 
 # estimator for initial step based on book
 # "Solving Ordinary Differential Equations I" by Hairer et al., p.169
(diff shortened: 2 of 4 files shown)


mauro3:
Note: I dropped Julia 0.3 support. Could be made to work but is a hassle. At this point dropping 0.3 should be fine.

mauro3:
I tidied this up and I think it is ready to be merged. As a second commit, I also snuck in an update to `.travis.yml` which avoids the red banner if it fails on nightly.  Both commits should not be controversial, so I'll merge them tomorrow if no-one complains.

The commit message says it all:

 Now unit tests are using Float32 and BigFloat

- Apart from the tests some small updates were needed in the solvers to
  make it all type stable.

- Issue in Julia with BigFloat and deepcopy:
  JuliaLang/julia#16667
  It was fixed for BigFloat in JuliaLang/julia#16999
  but not back-ported to Julia 0.4.  Thus not testing BigFloat on 0.4

- Fixed `@assert` -> `@test` in runtests.jl

- ODE.ode45_fe hangs with Float32, thus those tests are disabled.

ChrisRackauckas:
Looks good. I think `ode45_fe` might be hanging on 32-bit because the default absolute tolerance is 1e-8 and `eps(Float32)==1.19e-7`. Then again, I'm not sure why the other ones wouldn't stall because of this. I think it's fine to leave that be.

PR #99, opened 2016-06-06
AB adaptive

obiajulu:
Here we implement both an explicit and implicit adaptive Adam Bashforth solver, in addition to the fixed step Adam bashforth solver currently on the ob/a-b branch.


obiajulu:
Sorry, I meant to make that pull request against my local fork.

PR #101, opened 2016-06-10
Added DOPRI78 tableau from DifferentialEquations.jl

mauro3:
Note that I had to increase the test tolerance `tol` for the new solver to pass.  This is not due to an in-accurate solver but due to the dense output only being 3rd order.  This made this test fail:
https://github.com/JuliaLang/ODE.jl/blob/a0ef97d6249e62a089370787850ae697e3d94962/test/runtests.jl#L60

### DIFF
--- a/src/runge_kutta.jl
+++ b/src/runge_kutta.jl
@@ -20,7 +20,8 @@ immutable TableauRKExplicit{Name, S, T} <: Tableau{Name, S, T}
         @assert istril(a)
         @assert S==length(c)==size(a,1)==size(a,2)==size(b,2)
         @assert size(b,1)==length(order)
-        @assert norm(sum(a,2)-c'',Inf)<1e-10 # consistency.  
+        aa = T<:Rational ? convert(Matrix{Rational{BigInt}}, a) : a # to avoid overflow
+        @assert norm(sum(aa,2)-c'',Inf)<1e-10 # consistency.
         new(order,a,b,c)
     end
 end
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
The order of the dense output would render the test mostly moot. It's probably best to test at a calculated point (like the endpoint) since that would have a much higher accuracy for most of the methods.  Then the tolerance could be far reduced (or dependent on the method order)

mauro3:
Yes, the tests need some love but that should be coming with #49. I think it would be fine to mere this now, or close.  However, the tableau is not copied to #49 yet.

ChrisRackauckas:
Alright. I'll take that as a separate issue then, but the tests will need to be overhauled (I think some kind of more precise integration test than what's going on here would help). Is that handled in #49? If not, an issue should be opened discussing some ideas on how to improve the tests. 

But this can't merge until the conflicts are resolved. I'm kinda surprised that this would conflict with something. Is it just the BS2/3 line spacing change that conflicts with the just merged #97?

mauro3:
Ok, I'll update this. I'll see whether I got time today, afterwards I'm away from the computer for a 5 days.

pwl:
@ChrisRackauckas I encourage you to post an issue for #49, the new tests are pretty fresh and we any comments/improvements on them are welcome.  Also, dense output and integrators are now completely separated in #49, which means you can test an integrator without invoking dense output.

PR #102, opened 2016-06-10
Fixed to allow using matrices as scalar like entity

### LINKED ISSUE #86: Working on a matrix instead of vector
I'm working on a 2D diffusion equation with a single time derivative on the l.h.s. and x and y derivatives on the r.h.s:

d[ f(x,y,t) ]/dt = H f(x,y,t)

Without using ODE, I solve this with simply as

f(x,y,t+1) = f(x,y,t) + dt H f(x,y,t)

And it is more efficient to write f(x,y,t) as a matrix with x as lines as y as columns. So my function F(f) receives f as a matrix and returns another matrix out=H.f. In F(f) I work out the space derivatives using Fourier transforms.

This is not compatible with the ODE code, right? ODE requires f to be a vector.

Would it be difficult to generalize ODE to accept f(x,y,t) to be a matrix?

mauro3:
Fixes #86 

Note, this drops Julia 0.3 compatibility due to a bug in its `norm` function (shows up in `hinit`).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - osx
     - linux
 julia:
-    - 0.3
     - release
     - nightly
 git:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
-julia 0.3
+julia 0.4
 Polynomials
 Compat 0.4.1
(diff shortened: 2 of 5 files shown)


mauro3:
Would be great if someone could review this!  @obiajulu and @pwl are currently working on things which could use this (well at least the test case).

ChrisRackauckas:
This seems like a way of fitting everything into the Vector format (and forcing Vector{Vector} output), instead of directly handling AbstractArrays (and outputing Vector{AbstractArray}). What's the reason for this decision?

mauro3:
Yes, we (also @pwl) should discuss the best approach sometime soon.

PR #106, opened 2016-07-02
[WIP/RFC]:Addition of Variable and Fixed Adam Moulton Method

obiajulu:
# Overview of Pull Request

We add two new solvers for ODE.jl, and relocate the old multistep solver
ode_ms, namely ode_am and ode113. ode_am is a fixed step method similiar
to ode_ms, but uses a PECE step. ode113 is a variable step size and
variable order method using the same underlying theory as ode_am. For more
documentation, refer to Hairer et al Volume 1 on Adam Methods. We group all of the multistep solvers in a new file "adams_methods.jl", which meant moving `ode_ms` from "ODE.jl" to the new file. 
1. Implementation of `ode113`
2. Implementation of `ode_am`
3. Moving `ode_ms` to new file "adams_methods.jl"
# Performance results for `ode_am`

Using @mauro3's [IVPTestSuite.jl](github.com/mauro3/IVPTestSuite.jl), we present benchmarks which show the performace of both solvers. First, for `ode_am`.
### walltime vs scd for threebody

threebody_stepsizes
### walltime vs scd for Pleidas

plei_stepsizes
# Performance results for `ode113`

We also present results for `ode113` using IVPTestSuite.jl
### walltime vs scd for threebody

threebody_tols_from_10e-5_to_10e-14
### walltime vs scd for Pleidas

plei_tols_from_10e-5_to_10e-15
### function evaluations for Pleidas problem `ode113` vs `ode45`

The real strength of Adam Bashforth methods is the minimal amount of derivative function evaluation one needs. For each iteration step, this function needs to be evaluated usually only twice (really twice for each attempt of a step, and steps are usually accepted). We present a table of the function evaluations necessary for `ode45` vs `ode113` used to solve the same Pleiadas problem above.

| tol | ode113 | ode45 |
| --- | --- | --- |
| 1e-7 | 1115 | 2456 |
| 1e-8 | 1327 | 3890 |
| 1e-9 | 1557 | 6170 |
| 1e-10 | 1859 | 9776 |
| 1e-11 | 2237 | 15488 |
| 1e-12 | 2875 | 24548 |
| 1e-13 | 3869 | 38900 |
| 1e-14 | 6179 | 61652 |
### Notes
1. Dense output is given by a native method which has degree one less than the order the solver is using over that range. Thus, dense output should be better than the degree 3 hermite interpolation method.
2. Best not to push `reltol`/`abstol` less than ~10e-15
## 
# Work still to be done
- [x] Better handling of `make_consisent_types` 
- [x] Fix a small problem with backwards compatibility with v0.3 for array declaration in line 604, currently causing a failed Travis run for Julia v0.3
## 

I wanted to give thanks for @mauro3, @jiahao, and @yuyichao ingboMa for their help with development of these two solvers.

### DIFF
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -8,9 +8,9 @@ using Compat
 
 ## minimal function export list
 # adaptive non-stiff:
-export ode23, ode45, ode78
+export ode23, ode45, ode78, ode113
 # non-adaptive non-stiff:
-export ode4, ode4ms
+export ode4, ode4ms, ode4am
 # adaptive stiff:
 export ode23s
 # non-adaptive stiff:
@@ -93,7 +93,7 @@ end
 # of the allowed domain.  Used in adaptive step-control.
 isoutofdomain(x) = isnan(x)
 
-function make_consistent_types(fn, y0, tspan, btab::Tableau)
+function make_consistent_types(fn, y0, tspan)
     # There are a few types involved in a call to a ODE solver which
     # somehow need to be consistent:
     #
@@ -113,7 +113,6 @@ function make_consistent_types(fn, y0, tspan, btab::Tableau)
     # - Eyf: suitable eltype of y and f(t,y)
     #   --> both of these are set to typeof(y0[1]/(tspan[end]-tspan[1]))
     # - Ty: container type of y0
-    # - btab: tableau with entries converted to Et
 
     # Needed interface:
     # On components: /, -
@@ -138,6 +137,19 @@ function make_consistent_types(fn, y0, tspan, btab::Tableau)
     !isleaftype(Et) && warn("The eltype(tspan) is not a concrete type!  Change type of tspan for better performance.")
     !isleaftype(Eyf) && warn("The eltype(y0/tspan[1]) is not a concrete type!  Change type of y0 and/or tspan for better performance.")
 
+    return Et, Eyf, Ty
+end
+
+
+function make_consistent_types(fn, y0, tspan, btab::Tableau)
+    # Returns
+    # - Et: eltype of time, needs to be a real "continuous" type, at
+    #       the moment a AbstractFloat
+    # - Eyf: suitable eltype of y and f(t,y)
+    #   --> both of these are set to typeof(y0[1]/(tspan[end]-tspan[1]))
+    # - Ty: container type of y0
+    # - btab: tableau with entries converted to Et
+    Et, Eyf, Ty = make_consistent_types(fn,y0,tspan)
     btab_ = convert(Et, btab)
     return Et, Eyf, Ty, btab_
 end
@@ -147,48 +159,7 @@ end
 ###############################################################################
 
 include("runge_kutta.jl")
-
-# ODE_MS Fixed-step, fixed-order multi-step numerical method
-#   with Adams-Bashforth-Moulton coefficients
-function ode_ms(F, x0, tspan, order::Integer)
-    h = diff(tspan)
-    x = Array(typeof(x0), length(tspan))
-    x[1] = x0
-
-    if 1 <= order <= 4
-        b = ms_coefficients4
-    else
-        b = zeros(order, order)
-        b[1:4, 1:4] = ms_coefficients4
-        for s = 5:order
-            for j = 0:(s - 1)
-                # Assign in correct order for multiplication below
-                #  (a factor depending on j and s) .* (an integral of a polynomial with -(0:s), except -j, as roots)
-                p_int = polyint(poly(diagm(-[0:j - 1; j + 1:s - 1])))
-                b[s, s - j] = ((-1)^j / factorial(j)
-                               / factorial(s - 1 - j) * polyval(p_int, 1))
-            end
-        end
-    end
-
-    # TODO: use a better data structure here (should be an order-element circ buffer)
-    xdot = similar(x)
-    for i = 1:length(tspan)-1
-        # Need to run the first several steps at reduced order
-        steporder = min(i, order)
-        xdot[i] = F(tspan[i], x[i])
-
-        x[i+1] = x[i]
-        for j = 1:steporder
-            x[i+1] += h[i]*b[steporder, j]*xdot[i-(steporder-1) + (j-1)]
-        end
-    end
-    return vcat(tspan), x
-end
-
-# Use order 4 by default
-ode4ms(F, x0, tspan) = ode_ms(F, x0, tspan, 4)
-ode5ms(F, x0, tspan) = ODE.ode_ms(F, x0, tspan, 5)
+include("adams_methods.jl")
 
 ###############################################################################
 ## STIFF SOLVERS
@@ -418,11 +389,6 @@ ode4s_s(F, x0, tspan; jacobian=nothing) = oderosenbrock(F, x0, tspan, s4_coeffic
 
 # Use Shampine coefficients by default (matching Numerical Recipes)
 const ode4s = ode4s_s
-
-const ms_coefficients4 = [ 1      0      0     0
-                          -1/2    3/2    0     0
-                          5/12  -4/3  23/12 0
-                          -9/24   37/24 -59/24 55/24]
 
 
 end # module ODE
--- a/src/adams_methods.jl
+++ b/src/adams_methods.jl
@@ -1,0 +1,626 @@
+using ProgressMeter
+"""
+    t,y = ode_ms(odefun, y0, tspan, order=4) with tspan = [t0:dt:tfinal]
+
+Fixed-Equal Steps Adams-Bashforth multi-step solver for nonstiff problems
+
+(Main ref: Hairer & Wanner 1996, Vol I, p.359-360, 372)
+
+Note: If method is run at order k, the first step is taken at order 1,
+and the order increases each step until it is order k on the kth step.
+Thus results may be of lower accuracy than expected.
+"""
+
+function ode_ms(F, y0, tspan, order::Integer)
+    dt = diff(tspan)
+
+    y = Array(typeof(y0), length(tspan))
+    y[1] = y0
+
+    if 1 <= order <= 4
+        b = ms_coefficients4
+    else
+        b = zeros(order, order)
+        b[1:4, 1:4] = ms_coefficients4
+        for s = 5:order
+            for j = 0:(s - 1)
+                # Assign in correct order for multiplication below
+                #  (a factor depending on j and s) .* (an integral of a polynomial with -(0:s), except -j, as roots)
+                p_int = polyint(poly(diagm(-[0:j - 1; j + 1:s - 1])))
+                b[s, s - j] = ((-1)^j / factorial(j)
+                               / factorial(s - 1 - j) * polyval(p_int, 1))
+            end
+        end
+    end
+
+    # TODO: use a better data structure here (should be an order-element circ buffer)
+    dy = similar(y)
+    for i = 1:length(tspan)-1
+        # Need to run the first several steps at reduced order
+        steporder = min(i, order)
+        dy[i] = F(tspan[i], y[i])
+
+        # Check step size is equal or very near equal
+        if !isapprox(dt[i],dt[1])
+            error("Step sizes are not of constant length")
+        end
+
+        y[i+1] = y[i]
+        for j = 1:steporder
+            y[i+1] += dt[i]*b[steporder, j]*dy[i-(steporder-1) + (j-1)]
+        end
+    end
+    return vcat(tspan), y
+end
+
+# Use order 4 by default
+ode4ms(F, y0, tspan) = ODE.ode_ms(F, y0, tspan, 4)
+ode5ms(F, y0, tspan) = ODE.ode_ms(F, y0, tspan, 5)
+
+
+"""
+    t,y = ode_am((odefun, y0, tspan, order=4) with tspan = [t0:dt:tfinal]
+
+Fixed-Equal Step Adams-Moulton PECE solver for nonstiff problems
+
+(Main ref: Hairer & Wanner 1996, Vol I, p.359-360, 372)
+
+#Note: this function works for both scalar-like and array-like input
+
+#Note: If method is run at order k, the first step is taken at order 1,
+and the order increases each step until it is order k on the kth step.
+Thus results may be of lower accuracy than expected.
+"""
+
+##Function: ode_am(F,y0, t, order(optional))
+##order is set to 4 by default
+function ode_am(F,y0, t,order::Integer = 4)
+    if (1 <= order <= 4)
+        b_imp = am_imp_coefficients3
+        b_exp = ms_coefficients4
+    else
+        #calculating higher order coefficients for implicit Adams Moulton method
+        b_imp = zeros(order, order)
+        b_imp[1:4,1:4] = am_imp_coefficients3
+        k = order - 1 # For explicit method, order = k+1
+        for s = 4:k
+            for j = 0:s
+                # Assign in correct order for multiplication below
+                #  (a factor depending on j and s) .* (an integral of a polynomial with -(-1:s-1), except -(j-1), as roots)
+                p_int = Polynomials.polyint(Polynomials.poly(diagm(-[-1:j - 2; j:s-1])))
+                b_imp[s+1, s+1 - j] = ((-1)^j / factorial(j)
+                               / factorial(s - j) * Polynomials.polyval(p_int, 1))
+            end
+        end
+        b_exp = zeros(order, order)
+        b_exp[1:4, 1:4] = ms_coefficients4
+        k = order # For implicit method, order = k
+        for s = 5:k
+            for j = 0:(s - 1)
+                # Assign in correct order for multiplication below
+                #  (a factor depending on j and s) .* (an integral of a polynomial with -(0:s), except -j, as roots)
+                p_int = Polynomials.polyint(Polynomials.poly(diagm(-[0:j - 1; j + 1:s - 1])))
+                b_exp[s, s - j] = ((-1)^j / factorial(j)
+                               / factorial(s - 1 - j) * Polynomials.polyval(p_int, 1))
+            end
+        end
+    end
+
+    dt = diff(t)
+    y = Array(typeof(y0), length(t))
+    dy = similar(y)
+    y[1] = y0
+    dy[1] = F(t[1],y[1])
+
+    ##Predict-Evalulate-Correct-Evaluate (PECE) Method for Implicit Adams solver
+    for i=1:length(t)-1
+        steporder = min(i,order)
+
+        # Check step size is equal or very near equal
+        if !isapprox(dt[i],dt[1])
+            error("Step sizes are not of constant length")
+        end
+
+        ##(P)redict y[i+1] using explicit Adams Bashforth coefficients
+        y[i+1]=y[i]
+        for j=1:steporder
+            y[i+1] += dt[i]*b_exp[steporder, j]*dy[i-(steporder-1) + (j-1)]
+        end
+
+        ##(E)valuate function F at the approximate point t[i+1],y[i+1]
+        dy[i+1] = F(t[i+1],y[i+1])
+
+        ##(C)orrect the formula using implicit Adams Moulton coefficients
+        y[i+1]=y[i]
+        for j = 1 : steporder
+            y[i+1] += dt[i]*b_imp[steporder, j]*dy[i - (steporder -1) + j]
+        end
+
+        ##(E)valulate the function anew at corrected approximatiion t[i+1],y[i+1]
+        dy[i+1] = F(t[i+1], y[i+1])
+    end
+    return t, y
+end
+
+ode4am(F::Function,y0, t)= ode_am(F::Function,y0, t,4)
+ode5am(F::Function,y0, t)= ode_am(F::Function,y0, t,5)
+
+###############################################################################
+# TABLEAUS FOR ADAMS FIXED STEP SOLVERS
+###############################################################################
+
+##Adams Bashforth Coefficients for Explicit Method
+const ms_coefficients4 = [1        0        0         0
+                          -1//2    3//2     0         0
+                          5//12    -4//3    23//12    0
+                          -9//24   37//24   -59//24   55//24]
+
+##Adams Moulton Coefficients for Implicit Method
+const am_imp_coefficients3 =   [1         0        0         0
+                                1//2      1//2     0         0
+                                -1//12    8//12    5//12     0
+                                1//24     -5//24   19//24    9//24]
+
+
+const γ_star_array=[1.0, -0.5,-0.08333333333333334,-0.041666666666666664,
+                            -0.026388888888888885,-0.018750000000000003,
+                            -0.014269179894179893,-0.01136739417989418,
+                            -0.009356536596119928,-0.00789255401234568,
+                            -0.006785849984634706,-0.005924056412337663,
+                            -0.005236693257950285,-0.004677498407042264,
+                            -0.004214952239005473,-0.003826899553211884,
+                            -0.003497349845349918]
+"""
+#     t,y = ODE113(odefun, y0, tspan,order=4;kwords...)
+#
+#with tspan = [t0,tfinal] or[t0,t1,...,tfinal] for dense output
+#
+#Variable Step Variable Order (VSVO) Adams-Moulton Solver for nonstiff problems
+#
+#(Main ref: Hairer & Wanner 1996, Vol I, p.356-360, p.398-400, p.421-423)
+"""
+function ode113(F, y0, tspan; kwords...)
+    # For y0 which don't support indexing.
+    F_ = (t, y) -> [F(t, y[1])]
+    t,y = ode113(F_, [y0], tspan; kwords...)
+    return t, vcat_nosplat(y)
+end
+
+"a variable stepsize, variable order multi-step solver for nonstiff ODEs"
+function ode113(F, y0::AbstractVector, tspan; order::Integer = 4,
+                                        adaptive_order = true,
+                                        max_order = 13,
+                                        reltol = 1.0e-9,
+                                        abstol = reltol,
+                                        minstep = abs(tspan[1] - tspan[end])/1e18,
+                                        maxstep = abs(tspan[end] - tspan[1])/2.5,
+                                        initstep = 0.,
+                                        fac = .8,
+                                        norm = Base.norm,
+                                        points = :all,
+                                        progressmeter = false)
+
+    #-##########################################################################
+    # Initializations
+    #-##########################################################################
+    Et, Eyf, Ty = make_consistent_types(F, y0, tspan)
+    dof = length(y0)
+
+    # initialization of time span, which will have variable length
+    tspan = convert(Vector{Et}, tspan)
+    tstart = tspan[1]
+    tfinal = tspan[end]
+    t = [tstart]
+
+    # initialzation of y[n], which will have variable length as well
+    # TODO: fix after https://github.com/JuliaLang/julia/issues/16667
+    #y_n = deepcopy(y0)
+    y_n = copy(y0)
+
+    # initialzation of dy[n], which will have variable length as well
+    dy_n = F(t[1],y0)
+    Tdy = typeof(dy_n)
+
+    # Declare work variables for t_np1, y_np1 and dy_np1  which are for
+    # for t_{n+1}, y(t_{n+1}), and dy(t_{n+1})
+    t_np1   = tspan[1]
+    y_np1   = similar(y0, Eyf, dof)
+    dy_np1  = similar(dy_n)
+
+    # variables for dense output
+    tout = [tstart]
+    iter = 2
+    yout = Array(Ty, 1)
+    yout[1] = copy(y0)
+
+    #initialization of step size dt, which will have variable length as well
+    dt = initstep # current trial stepsize for a give step
+    prev_dt = dt # the stepsize of the last successful step
+    next_dt = 0.0 # stepsize to be used in next step
+
+    if dt == 0.
+        # initial guess at a step size
+        # dt, tdir, dy0 = hinit(F, y0, tstart, tfinal, 3, reltol, abstol)
+        dt, tdir, dy0 = hinit(F, y0, tstart, tfinal, 1, reltol, abstol)
+        dt = dt
+    else
+        tdir = sign(tfinal - tstart)
+    end
+
+    n = 1     #`n` counts the time steps. We start at n=1, the first step
+    num_stages = 0 # `num_stages`+1 is the order of the AB method at the current step
+                   # Note: that due to Julia starting indexing at 1, whereas Hairer
+                   # indices his variable for order, k, from 0, we have
+                   # k= num_stages in what follows
+
+    # initialization of arrays used in estimation y_np1 given dy[n],y[n], dt and
+    # list order timestep are  initialized to a maxorder x maxorder dimension
+    c,g,ϕstar_n, ϕ_n, ϕstar_nm1, b, ϕ_np1 = intialize_coef_arrays(max_order+3,Tdy,Et)
+
+    if max_order <= 13
+        γ_star = γ_star_array
+    else
+        γ_star= gamma_star_constr(max_order+3)
+    end
+
+    # Declaration of timeout variables
+    timeout_VS_const = 0 # How many steps before a variable step size can be taken
+    timeout_VS = timeout_VS_const # `timeout_VS` is a counter for this in the loop
+
+    timeout_VO_const = 2 # How many steps before a variable step order can be taken
+    timeout_VO = timeout_VO_const # `timeout_VO` is a counter for this in the loop
+
+    consec_const_steps = 0 # counts how many consecutives steps have been taken
+                           # with same step size
+
+    # Declaration of  four local error (LE) or truncation errors necessary
+    # for variable order calculations, which correspond to infinity norms of:
+    #      LE(n+1)_{k-1}, LE(n+1)_{k}, E(n+1)_{k+1}, and LE(n+1)_{k+2}
+    # from Hairer pgs 422-424
+    err_km1, err_k, err_kp1, err_kp2 = 0.0,0.0,0.0,0.0
+
+    #-##########################################################################
+    # Variable Step Variable Order (VSVO) Step Integration Loop
+    #-##########################################################################
+    progressmeter
+    if progressmeter
+        prog = Progress(round(Int,1000*tdir*(tfinal-tstart)), "Solving ODE:")
+    end
+    while  tfinal*tdir > t[n]*tdir
+        #--#####################################################################
+        # Attempt a step with current trial stepsize `dt`
+        #--#####################################################################
+        successful_step = false
+        while !successful_step
+            #---################################################################
+            # Make an approriate time step with current trial stepsize `dt`
+            #---################################################################
+            #Make the step
+            t_np1 = t[n]+dt
+            #---################################################################
+            # Calculating explicit approximatiion to y_np1 with current step dt
+            #---################################################################
+
+            # Calculation of g_{j}(n)coefficients for j=0,...,k+1;
+            g_coefs!(g,c,dt,t,t_np1,n,min(n,(num_stages+1)+1))
+            #@show min(max(n-1,1),(num_stages)+1)
+            # Calculation of  ϕstar_{j}(n), and ϕ_{j}(n) coef. for j=0,...,k;
+            #ϕ_slots = min(max(n-1,1),(num_stages)+1)
+            ϕ_and_ϕstar_coefs!(ϕstar_n, ϕ_n, ϕstar_nm1, b,dy_n,t,t_np1,n,1:(num_stages)+1)
+            #Note that  1<= min(max(n-1,1),num_stages+1) <= n-1 and num_stages+1
+
+            # Calculate predictor with explicit method of order `num_stages`. See
+            # equation (5.5) in Hairer et al, Vol 1, where we note that here
+            # y_np1 = y_{k}(t_{n+1})
+
+            copy!(y_np1,y_n)
+            for j= 0:num_stages-1
+                y_np1 += dt*g[(j)+1]*ϕstar_n[(j)+1]
+            end
+
+            # Compute current estimate for derivative
+            dy_np1=F(t_np1,y_np1)
+
+            #---################################################################
+            # Correction-Evaluaion (CE) Loop for implicit solver
+            #     See Hairer p. 360 for a general description of this method
+            #---################################################################
+
+            # Set CE_loops to be how many correction-evaulation cycles
+            # desired. As least one loop is necessary, for the sake of the
+            # adaptive step selection
+            CE_loops = 1
+            for loop_counter=1:CE_loops
+                # (C)orrection of y_np1
+                # Add new coefficients  based on this prediction
+                # computation of ϕ, ϕstar using recurrence relations
+                ϕ_coefs!(ϕ_np1, ϕstar_n,dy_np1,((0)+1):((num_stages)+1))
+
+                # Calculation of g_{num_stages + 1}(n) coefficients using
+                # recurrence relations has already been done above, and does
+                # not depend on dy_np1 or y_np1, so no update necessary
+
+                # We correct the computation of y_np1 using y_np1
+                # as part of our prediction Hairer equation 5.7
+                y_np1 += dt*g[(num_stages)+1]*ϕ_np1[(num_stages)+1]
+
+                # (E)valuate estimate for next derivative again
+                dy_np1=F(t_np1,y_np1)
+            end
+            #---################################################################
+            # Error Check for step size `dt` and Variable next step selection:
+            #     Check error to determine if step size should be accepted:
+            #     ->if yes, compute `next_dt` for next step
+            #     ->if no, use this `next_dt`` for next attempt at this step
+            #---################################################################
+
+            #Don't allow increases of steps until n >= num_stages + 2
+            if n==1
+                successful_step = true
+                next_dt = dt/4
+            else
+                # We calculate local truncation error, LE_{k+1}; the difference
+                # between implicit results for order k and k+1.
+
+                # We need ϕ_{k+1}(n+1) and g_{k+1}(n);the latter has already
+                # been calculated above, the former still needs calculation
+                #@show n, num_stages
+                ϕ_coefs!(ϕ_np1, ϕstar_n,dy_np1,(num_stages+1)+1)
+
+                # Calculate ||LE_{k+1}|| and next stepsize
+
+                yerr_kp1= dt*(g[(num_stages+1)+1] - g[(num_stages)+1])*ϕ_np1[(num_stages+1)+1]
+                err_kp1, next_dt, timeout_VS = stepsize_hw92!(dt,tdir,y_n,y_np1,
+                                            yerr_kp1, num_stages, timeout_VS,dof,
+                                            abstol, reltol, maxstep, norm)
+
+                # Hairer equation 7.4: If ||LE_{k+1}|| =< 1
+                # then step size is accepted and use next_dt for next step
+                # Else, then step size is rejected, reattempt with fac*next_dt
+                if err_kp1 <= 1
+                    successful_step = true
+                else
+                    successful_step = false
+                    dt = fac*next_dt
+                end
+            end # End of error check for current step size dt
+        end # End of loop for finding successfull step size for nth step
+            # repeats if success_step == false
+
+        #Keep step size constant if `dt` is close to `prev_dt`
+        if abs(next_dt/prev_dt)<2 && abs(next_dt/prev_dt)>=1
+            consec_const_steps += 1
+            next_dt = prev_dt
+        else
+            consec_const_steps = 0
+        end
+
+        #Don't step past `tfinal`
+        if tdir*(t_np1 + next_dt) >= tdir*(tfinal)
+            next_dt = tfinal-t_np1
+        end
+        push!(t,copy(t_np1))
+
+        #--#####################################################################
+        # Interpolation for dense output
+        #--#####################################################################
+
+        # interpolate specified points
+        # TODO: less expensive interpolation. Core problem is that current
+        # frame work does not allow for inexpensive interpolation for uneven
+        # step sizes.
+
+        while iter <= length(tspan) && tdir*tspan[iter]<tdir*t_np1
+            dt_iter = tspan[iter]-t[n]
+
+            # Calculation of g_{j}(n)coefficients for j=0,...,k+1;
+            g_coefs!(g,c,dt,t,tspan[iter],n,min(n,num_stages+1))
+
+            # Calculation of  ϕstar_{j}(n), and ϕ_{j}(n) coef. for j=0,...,k;
+            ϕ_and_ϕstar_coefs!(ϕstar_n, ϕ_n, ϕstar_nm1, b,dy_n,t,tspan[iter],n,1:min(max(n-1,1),num_stages+1))
+
+            yout_iter = copy(y_n)
+            for j = 0:num_stages-1
+                yout_iter += dt_iter*g[(j)+1]*ϕstar_n[(j)+1]
+            end
+
+            # CE iteration for more precise interpolation
+            if n>=num_stages+1
+                dy_iter=F(tspan[iter],yout_iter)
+                ϕ_coefs!(ϕ_np1, ϕstar_n,dy_iter,1:(num_stages)+1)
+
+                yout_iter += dt_iter*g[(num_stages)+1]*ϕ_np1[(num_stages)+1]
+            end
+
+            push!(tout,tspan[iter])
+            push!(yout, copy(yout_iter))
+            iter += 1
+        end
+
+        if iter <= length(tspan) && tspan[iter] == t_np1
+            # handling when tspan and choosen time steps overlap
+            push!(tout, copy(t_np1))
+            push!(yout, copy(y_np1))
+            iter += 1
+        elseif points == :all
+            push!(tout, copy(t_np1))
+            push!(yout, copy(y_np1))
+        end
+
+        # restore values of coefficients for choosen step size
+        g_coefs!(g,c,dt,t,t_np1,n,min(n,(num_stages)+1))
+        ϕ_and_ϕstar_coefs!(ϕstar_n, ϕ_n, ϕstar_nm1, b, dy_n,t,t_np1,n,1:((num_stages)+1))
+        ϕ_coefs!(ϕ_np1, ϕstar_n,dy_np1,1:((num_stages)+1))
+
+        #--#####################################################################
+        # Variable order selection for next step
+        #   Not run until a success step was taken for n-th step
+        #   See Hairer pg 424 for indepeth description
+        #--#####################################################################
+
+        current_stageorder = num_stages
+        if n<=order-1
+            # Need to run the first few steps at reduced order for it requires
+            # the last 'order' number of points
+            num_stages = min(n, max_order)
+        elseif adaptive_order && num_stages>1
+            # First, compute truncation errors for implicit orders: k-1,k
+            # against current k+1
+
+            yerr_km1 = dt*(g[(num_stages-1)+1] - g[(num_stages-2)+1])*ϕ_np1[(num_stages-1)+2]
+            err_km1 = stepsize_hw92!(dt, tdir, y_n, y_np1, yerr_km1, num_stages,
+                            timeout_VS,dof, abstol, reltol, maxstep, norm)[1]
+
+            yerr_k = dt*(g[(num_stages)+1] - g[(num_stages-1)+1])*ϕ_np1[(num_stages)+1]
+            err_k = stepsize_hw92!(dt, tdir, y_n, y_np1, yerr_k, num_stages,
+                            timeout_VS, dof, abstol, reltol, maxstep, norm)[1]
+
+            # Decrement num_stages if LE_{k} and LE_{k-1} < LE_{k+1}
+            if max(err_km1,err_k) <= err_kp1
+                num_stages = max(order,num_stages - 1)
+
+            elseif consec_const_steps >= num_stages && timeout_VO==0 && n>(num_stages + 1)+1
+                # Given enough consecutive steps taken at a constant step size
+                # we compute truncation error for order implicit k+2
+                ϕ_and_ϕstar_coefs!(ϕstar_n, ϕ_n, ϕstar_nm1, b, dy_n,t,t_np1,n,((num_stages+1)+1))
+                ϕ_coefs!(ϕ_np1, ϕstar_n,dy_np1,(num_stages+2)+1)
+
+                yerr_kp2 = dt*γ_star[(num_stages + 2)+1]*ϕ_np1[(num_stages+2)+1]
+                err_kp2 = stepsize_hw92!(dt,tdir,y_n,y_np1,yerr_kp2,num_stages,
+                            timeout_VS, dof, abstol, reltol, maxstep, norm)[1]
+
+                # Increment num_stages if LE_{k+2} < LE_{k+1} and max(LE_{k},
+                # LE_{k-1}) > LE_{k+1}
+                if err_kp2 < err_kp1
+                    num_stages = min(max_order,num_stages + 1)
+                    timeout_VO = timeout_VO_const # reset timout_VO
+                end
+            end
+            # Keep timeout_VO 0, or decrement
+            timeout_VO= timeout_VO==0? 0 : timeout_VO-1
+        end # end of adaptive order selection
+
+        #--#####################################################################
+        # Prepare for next step
+        #--#####################################################################
+
+        # Shift data from this step, from the "current step" position to the
+        # "previous step" position for ϕstar and ϕ
+        # Note: ϕ_n[:] and ϕ_star[2,:] corresponds to ϕ_n and ϕ*_n, and
+        # ϕ_star[1,:] to ϕ*_{n-1}, etc
+
+        ϕstar_nm1 = copy(ϕstar_n)
+        #ϕstar_nm1[1:min(max(n-1,1),current_stageorder+1)] =
+        #                        ϕstar_n[1:min(max(n-1,1),current_stageorder+1)]
+
+        # Update stepsize variables
+        prev_dt = dt
+        dt = next_dt
+        # push work variables
+        y_n = y_np1
+        dy_n = dy_np1
+
+        # Increment our step
+        n= n+1
+        if progressmeter
+          prog.counter = max(1,round(Int,1000.0*tdir*(t[n]-tstart)))
+          ProgressMeter.updateProgress!(prog; showvalues = [(:steps,n),(:current_t,t[n]),(:tfinal,tfinal)])
+        end
+    end #while loop over steps
+
+    return vcat(tout), yout
+end
+
+###############################################################################
+#
+# HELPER FUNCTIONS FOR ADAMS VSVO SOLVER
+#
+###############################################################################
+
+###############################################################################
+# Gamma star function from Hairer et al pg 359, table 1.2
+# used in adaptive step selection (See Hairer pg 423)
+###############################################################################
+#precompute factorial_values
+"compute γ*(j) for higher order versions of ode113"
+function gamma_star_constr(index)
+    γ_star = zeros(index)
+    #γ_star[1] = 1
+    for j = 1:index
+        γ_star[j] = Polynomials.polyval((1 / factorial(j))*
+            Polynomials.polyint(Polynomials.poly(diagm(collect(1-(j-1):1)))),1)
+    end
+    return γ_star
+end
+################################################################################
+# Helper functions for calculating coefficients from recurence relationships
+################################################################################
+#Based on Hairer Lemma 5.1 on page 399
+#Calculates c_j,q and g_n,j for n and j =1:num_stages
+"compute c_{j,q} and g_{n,j} coefficients for ode113"
+function g_coefs!(g,c,dt,t,t_np1,n,num_stages)
+    for j = 1:num_stages
+      for q = 1:((num_stages)-(j-1))
+        if j ==1
+          c[j,q]=1/q
+        elseif j==2
+          c[j,q]=1/q/(q+1)
+        else
+          c[j,q] = c[j-1,q] - c[j-1,q+1]*dt/(t_np1-t[n-(j-1)+1])
+        end
+      end
+      g[j]=c[j,1]
+    end
+    return nothing
+end
+
+
+#Based on Hairer equation 5.9. Calculates b, ϕ, ϕstar for
+#n and j  in range_of_indices
+"compute ϕ_{n,j} and ϕ*_{n,j} coefficients for ode113"
+function ϕ_and_ϕstar_coefs!(ϕstar_n, ϕ_n, ϕstar_nm1, b,dy_n,t,t_np1,n,range_of_indices)
+    for j in range_of_indices #0:k-1
+      if j== 1
+        b[j] = 1
+        ϕ_n[j] = copy(dy_n)
+        ϕstar_n[j] = copy(dy_n)
+      else
+        b[j] = b[j-1]*(t_np1-t[n-(j-1)+1])/(t[n]-t[n-(j-1)])
+        ϕ_n[j] =  ϕ_n[j-1]- ϕstar_nm1[j-1]
+        ϕstar_n[j] = b[j]*ϕ_n[j]
+      end
+    end
+    return nothing
+end
+
+
+#Based on Hairer equation 5.9. Just calculates the ϕ coefficients
+"compute ϕ_{n+1,j} coefficients for ode113"
+#ϕ_np1 as first
+function ϕ_coefs!(ϕ_np1, ϕstar_n,dy_n,range_of_indices)
+    for j  in range_of_indices
+      if j== 1#0
+          ϕ_np1[j] = copy(dy_n)
+      else
+        #Note: since n>=j, when this runs, n>=2 (so n-1>=1) and
+        #@show ϕstar_n[j-1]
+        ϕ_np1[j] = ϕ_np1[j-1] -  ϕstar_n[j-1]
+      end
+    end
+    return nothing
+end
+
+################################################################################
+# Helper functions initializing/resizing arrays
+################################################################################
+"used to initialize working arrays used in ode1113"
+function intialize_coef_arrays(max_order,T,Et)
+    c = zeros(Et,max_order,max_order)
+    g = zeros(max_order)
+    b = zeros(max_order)
+    ϕstar_nm1 = Array{T}(1,max_order)
+    ϕstar_n = Array{T}(1,max_order)
+    ϕ_n = Array{T}(1,max_order)
+    ϕ_np1 = Array{T}(1,max_order)
+    return c,g,ϕstar_n, ϕ_n, ϕstar_nm1, b, ϕ_np1
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,6 +1,5 @@
 using ODE
 using Base.Test
-
 tol = 1e-2
 
 solvers = [
@@ -12,12 +11,14 @@ solvers = [
            ODE.ode4,
            ODE.ode4ms,
            ODE.ode5ms,
+           ODE.ode4am,
            # adaptive
 #           ODE.ode21, # this fails on Travis with 0.4?! TODO revert once fixed.
            ODE.ode23,
            ODE.ode45_dp,
            ODE.ode45_fe,
            ODE.ode78,
+           ODE.ode113,
 
            ## Stiff
            # fixed-step
--- a/test/interface-tests.jl
+++ b/test/interface-tests.jl
@@ -45,19 +45,19 @@ ODE.isoutofdomain(y::CompSol) = any(isnan, vcat(y.rho[:], y.x, y.p))
 
 
 ################################################################################
- 
+
 # define RHSs of differential equations
 # delta, V and g are parameters
 function rhs(t, y, delta, V, g)
   H = [[-delta/2 V]; [V delta/2]]
- 
+
   rho_dot = -im*H*y.rho + im*y.rho*H
   x_dot = y.p
   p_dot = -y.x
- 
+
   return CompSol( rho_dot, x_dot, p_dot)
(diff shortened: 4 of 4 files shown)


src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+#
+#     [T,X] = ode_ms(ODEFUN, Y0, TSPAN,ORDER=4) with TSPAN = [T0:H:TFINAL]
+#
+#     Note: If method is run at order k, the first step is taken at order 1,
+#     and the order increases each step until it is order k on the kth step.
+#     Thus results may be of lower accuracy than expected.
+#     (Main ref: Hairer & Wanner 1996, Vol I, p.359-360, 372)
+################################################################################
+
+function ode_ms(F, y0, tspan, order::Integer)
```
mauro3: I think this solver only works when all `h` are equal, correct?  If so you should add test & error for that.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+#     Note: If method is run at order k, the first step is taken at order 1,
+#     and the order increases each step until it is order k on the kth step.
+#     Thus results may be of lower accuracy than expected.
+#     (Main ref: Hairer & Wanner 1996, Vol I, p.359-360, 372)
+################################################################################
+
+
+##Function: ode_am(F,y0, t, order(optional))
+##order is set to 4 by default
+function ode_am(F::Function,y0, t,order::Integer = 4)
```
mauro3: Does this function work with both scalar-like and array-like?  Maybe note this?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+end
+
+# Use order 4 by default
+ode4ms(F, y0, tspan) = ODE.ode_ms(F, y0, tspan, 4)
+ode5ms(F, y0, tspan) = ODE.ode_ms(F, y0, tspan, 5)
+
+
+################################################################################
+#  Fixed Step Adams-Moulton PECE solver for nonstiff problems
+#
```
mauro3: `PECE` either remove or define.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+#     [T,X] = ODE_AM(ODEFUN, Y0, TSPAN,ORDER=4) with TSPAN = [T0:H:TFINAL]
+#
+#     Note: If method is run at order k, the first step is taken at order 1,
+#     and the order increases each step until it is order k on the kth step.
+#     Thus results may be of lower accuracy than expected.
+#     (Main ref: Hairer & Wanner 1996, Vol I, p.359-360, 372)
```
mauro3: Maybe say: "A possible solution to this is to take very small first steps 1:order."

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            y[i+1] += dt[i]*b_imp[steporder, j]*dy[i - (steporder -1) + j]
+        end
+
+        ##(E)valulate the function anew at corrected approximatiion t[i+1],y[i+1]
+        dy[i+1] = F(t[i+1], y[i+1])
+    end
+    return t, y
+end
+
+ode_am5(F::Function,y0, t)= ode_am(F::Function,y0, t,5)
```
mauro3: Should this be `ode5ms`?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    end
+end
+
+###############################################################################
+#
+# HELPER FUNCTIONS and TABLEAUS FOR ADAMS SOLVERS
+#
+###############################################################################
+##Adams Bashforth Coefficients for Explicit Method
+const ms_coefficients4 = Float64[1        0        0         0
```
mauro3: These shouldn't be `Float64`.  This will result in a loss of accuracy when using something like `BigFloat`

src/ODE.jl
```diff
@@ -8,9 +8,9 @@ using Compat
 
 ## minimal function export list
 # adaptive non-stiff:
-export ode23, ode45, ode78
+export ode23, ode45, ode78, ode113
 # non-adaptive non-stiff:
-export ode4, ode4ms
+export ode4, ode4ms, ode_am4
```
pwl: How about we don't export any symbols and make the `ODE.odeXX` convention a standard?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+#     (Main ref: Hairer & Wanner 1996, Vol I, p.356-360, p.398-400, p.421-423)
+################################################################################
+function ode113(F, y0, tspan, order::Integer=4; kwords...)
+    # For y0 which don't support indexing.
+    F_ = (t, y) -> [F(t, y[1])]
+    t,y = ode113(F_, [y0], tspan, order; kwords...)
+    return t, vcat_nosplat(y)
+end
+
+"a variable stepsize, variable order multi-step solver for nonstiff ODEs"
```
mauro3: Either do doc-strings properly or not at all.  I think it is ok to do the latter, as no methods have doc-strings so far.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+function ode113(F, y0::AbstractVector, tspan, order::Integer=4;adaptive_order = true,
```
mauro3: Newline after `;`

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+                                                      max_order = 13,
+                                                      reltol = 1.0e-9,
+                                                      abstol = 1.0e-9,
+                                                      minstep=abs(tspan[1] - tspan[end])/1e18,
+                                                      maxstep=abs(tspan[end] - tspan[1])/2.5,
+                                                      initstep=0.,
+                                                      facmax = 1.5,
```
mauro3: Whilst I agree that those should become options at some point, probably shouldn't introduce new keywords in this PR.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    t_np1   = tspan[1]
+    y_np1   = similar(y0, Eyf, dof)
+    dy_np1  = similar(dy[1])
+
+    # variables for dense output
+
+    tout = [tstart]
+    iter = 2
+    yout = Array(Ty, 1)
+    allocate!(yout, y0, dof)
```
mauro3: This is pointless.  First, `allocate!` only makes sense to use with large vectors, not just length 1.  Then in line 199 you assign a copy of y0 to this slot, so the allocation is never used anyway.  199 would need to read `yout[1][:] = copy(y0)` or better `copy!(yout[1],y0)`.  However, these shenanigans will have no performance impact, instead write `yout = [copy(y0)]`.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    dof = length(y0)
+
+    # initialization of time span, which will have variable length
+    tspan = convert(Vector{Et}, tspan)
+    tstart = tspan[1]
+    tfinal = tspan[end]
+    t = [tstart]
+
+    # initialzation of y[n], which will have variable length as well
+    y = Array(Ty, 1)
```
mauro3: See comment below on `yout`.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    allocate!(y, y0, dof)
+    # TODO: fix after https://github.com/JuliaLang/julia/issues/16667
+    # y[1] = deepcopy(y0)
+    y[1] = copy(y0)
+
+    # initialzation of dy[n], which will have variable length as well
+    Tdy = typeof(F(t[1], y[1]))
```
mauro3: I'd write:
```
_dy = F(t[1],y[1])
Tdy = typeof(_dy)
dy = Tdy[_dy] 
```

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    dy = Tdy[]
+    push!(dy,F(t[1], y[1]))
+
+    # preallocate work variables for t_np1, y_np1 and dy_np1  which are for
+    # for t_{n+1}, y(t_{n+1}), and dy(t_{n+1})
+    t_np1   = tspan[1]
```
mauro3: There is no need to allocate a scalar and it cannot be done anyway (well, there are tricks but they do not improve performance). Plus line 266 creates a new binding anyway and does not update the value of `t_np1` in-place.  I think you're confused between values and bindings, read http://www.johnmyleswhite.com/notebook/2014/09/06/values-vs-bindings-the-map-is-not-the-territory/

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    y_np1   = similar(y0, Eyf, dof)
+    dy_np1  = similar(dy[1])
+
```
mauro3: Conversely, it is good to allocate work-arrays.  However, below you're making the mistake of re-binding them.  I.e. the preallocated memory is not used after all!  See comments below.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            # Calculate next value, y_np1, with order = k = steporder + 1. See
+            # equation (5.5) in Hairer et al, Vol 1, where we note that here
+            # y_np1 = y_{k}(t_{n+1})
+
+            y_np1 = y[n]
+            for j= 1:steporder
+              y_np1 += h*g[j]* ϕstar[2,j]
+            end
+            # TODO: figure out why the following does not work
```
mauro3: I see your TODO here.  If this
```
            copy!(y_np1, y[n])
            for j= 1:steporder
                for i=1:length(y_np1)
                    y_np1 += h*g[j]* ϕstar[2,j]
                end
            end
```
does not work, then I suspect there is a bug somewhere.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            # for j= 1:steporder
+            #     for i=1:length(y_np1)
+            #         y_np1[i] += h*g[j]*ϕstar[2,j][i]
+            #     end
+            # end
+
+            # Compute current estimate for derivative
+            dy_np1=F(t_np1,y_np1)
```
mauro3: As `F` is not in-place it will create a new vector anyway.  Thus it is pointless to then copy this into another memory slot.  Thus delete `dy_np1` in the initialization on line 190: https://github.com/JuliaLang/ODE.jl/pull/106/files#diff-43ccd017e404a50f115b60e06e7f1ac5R190

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+
+    tout = [tstart]
+    iter = 2
+    yout = Array(Ty, 1)
+    allocate!(yout, y0, dof)
+    # TODO: see line 177
+    yout[1] = copy(y0)
+
+    #initialization of step size h, which will have variable length as well
+    h = initstep # current trial stepsize for a give step
```
mauro3: Convention in ODE.jl is converging towards using `y,t,dt`.  Thus use `dt` instead of `h`.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+
+    else
+        tdir = sign(tfinal - tstart)
+    end
+
+    n = 1     #`n` counts the time steps. We start at n=1, the first step
+    steporder=1 # `steporder` is the order of the AB method at the current step
+                # Note: that due to Julia starting indexing at 1, whereas Hairer
+                # indices his variable for order, k, from 0, we have
+                # k = steporder + 1 in what follows
```
mauro3: shouldn't this be `k+1=steporder`?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+
+    # initialization of arrays used in estimation y_np1 given dy[n],y[n], h and
+    # list order timestep are  initialized to a maxorder x maxorder dimmension
+    c,g,b,ϕstar, ϕ = intialize_coef_arrays(max_order+3,Tdy)
```
mauro3: dimension

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    timeout_VS_const = 0 # How many steps before a variable step size can be taken
+    timeout_VS = timeout_VS_const # `timeout_VS` is a counter for this in the loop
+
+    timeout_VO_const = 2 # How many steps before a variable step order can be taken
+    timeout_VO = timeout_VO_const # `timeout_VO` is a counter for this in the loop
+
+    consec_const_steps = 0 # counts how many consecutives steps have been taken
+                           # with same step size
+
+    # The four local error (LE) or truncation error necessary for variable
```
mauro3: Again, binding these to 0.0 does nothing as you're re-binding later.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+      end
+    end
+end
+
+################################################################################
+# Helper functions initializing/resizing arrays
+################################################################################
+"used to initialize working arrays used in ode1113"
+function intialize_coef_arrays(dim,T)
+    c = zeros(dim,dim)
```
mauro3: Probably `zeros(T,dim,dim)`?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            if abs(h/prev_h)<2 && abs(h/prev_h)>=1
+                consec_const_steps += 1
+                h = prev_h
+            else
+                consec_const_steps = 0
+            end
+
+            #Don't step past `tfinal`
+            if tdir*(t[n] + h) >= tdir*(tfinal)
+                h = tfinal-t[n]
```
mauro3: Shouldn't `consec_const_steps` be updated here too?  Maybe it would be better to do this check when calculating the new `h`?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+        if j ==1
+          c[j,q]=1/q
+        elseif j==2
+          c[j,q]=1/q/(q+1)
+        else
+          c[j,q] = c[j-1,q] - c[j-1,q+1]*h/(t_np1-t[n-(j-1)+1])
+        end
+      end
+      g[j]=c[j,1]
+    end
```
mauro3: I like to have an explicit `return nothing` statement in functions like this and the ones below.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+        error("order must be positive integer")
+    end
+end
+
+################################################################################
+# Helper functions for calculating coefficients from recurence relationships
+################################################################################
+#Based on Harier Lemma 5.1 on page 399
+#Calculates c_j,q and g_n,j for n and j =1:steporder
+"compute g_{n,j} coefficients for ode113"
```
mauro3: But this computes the `c` coefs too.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    for j  in range_of_indices
+      if j== 1#0
+        ϕ[3,j] = dy_n
+      else j>1#0
+        #Note: since n>=j, when this runs, n>=2 (so n-1>=1) and
+        ϕ[3,j] =  ϕ[3,j-1]- ϕstar[2,j-1]
+      end
+    end
+end
+
```
mauro3: What is in `\phi[1,:]`?  It is written to in line https://github.com/JuliaLang/ODE.jl/pull/106/files#diff-43ccd017e404a50f115b60e06e7f1ac5R490, but I cannot see it used.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            #     See Harier p. 360 for a general description of this method
+            #---################################################################
+
+            # Set CE_loops to be how many correction-evaulation cycles
+            # desired. As least one loop is necessary, for the sake of the
+            # adaptive step selection
+            CE_loops = 1
+            loop_counter = 0
+
+            while loop_counter<CE_loops && n>=steporder+1
```
mauro3: I think writing it as
```
if n>=steporder+1
    for loop_counter=1:CE_loops
...
```
is clearer.  (Plus it avoids some checks)

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            #     Check error to determine if step size should be accepted:
+            #     ->if yes, compute `next_h` for next step
+            #     ->if no, use this `next_h`` for next attempt at this step
+            #---################################################################
+
+            #Don't use adaptive steps until n >= steporder + 2
+            if n==1
+                #TODO: adaptive step selection for n<steporder +2
+                successful_step = true
+                next_h = h/4
```
mauro3: I don't understand this `next_h = h/4`, why reduce the step size for the second step?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            elseif n<steporder+2
+                yerr_k = h*(g[steporder+1] - g[steporder])*ϕ[3,steporder+1]
```
mauro3: This should be in-place too.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            else
+                # We calculate local truncation error, LE_{k+1}; the difference
+                # between implicit results for order k and k+1.
+
+                # We need ϕ_{k+1}(n+1) and g_{k+1}(n);the latter has already
+                # been calculated above, the former still needs calculation
+                ϕ_coefs!(ϕ, ϕstar,dy_np1,steporder+2)
+
+                # Calculate ||LE_{k+1}|| and next stepsize
+                yerr_kp1 = h*(g[steporder+2] - g[steporder+1])*ϕ[3,steporder+2]
```
mauro3: in-place?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+                ϕ_coefs!(ϕ, ϕstar,dy_np1,1:steporder+1)
+
+                # Calculation of g_{steporder + 1}(n) coefficients using
+                # recurrence relations has already been done above, and does
+                # not depend on dy_np1 or y_np1, so no update necessary
+
+                # We correct the computation of y_np1 using y_np1
+                # as part of our prediction Harier equation 5.7
+
+                y_np1 = y_np1 + h*g[steporder+1]*ϕ[3,steporder+1]
```
mauro3: Yes, that TODO needs fixing.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+                dy_np1=F(t_np1,y_np1)
+            end
+            #---################################################################
+            # Error Check for step size `h` and Variable next step selection:
+            #     Check error to determine if step size should be accepted:
+            #     ->if yes, compute `next_h` for next step
+            #     ->if no, use this `next_h`` for next attempt at this step
+            #---################################################################
+
+            #Don't use adaptive steps until n >= steporder + 2
```
mauro3: If I read below correctly, it means that don't increase step-size until n >= steporder + 2.  But decreases are allowed after the first step.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+                next_h = h/4
+            elseif n<steporder+2
+                yerr_k = h*(g[steporder+1] - g[steporder])*ϕ[3,steporder+1]
+                err_k, next_h,timeout_VS = stepsize_hw92!(h, tdir, y[n], y_np1, yerr_k, steporder,
+                                timeout_VS, dof, abstol, reltol, maxstep, norm)
+                if err_k <= 1
+                    successful_step = true
+                else
+                    successful_step = false
+                    h = fac*next_h # make sure fac < 1
```
mauro3: If I understand correctly, `fac` should not be used here, it scales the estimated stepsize in `stepsize_hw92!` to something on the safe side.  I don't think this applies here.  Maybe 0.5` or `0.25` instead?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+                                            abstol, reltol, maxstep, norm)
+
+                # Harier equation 7.4: If ||LE_{k+1}|| =< 1
+                # then step size is accepted and use next_h for next step
+                # Else, then step size is rejected, reattempt with fac*next_h
+                if err_kp1 <= 1
+                    successful_step = true
+                else
+                    successful_step = false
+                    h = fac*next_h # make sure fac < 1
```
mauro3: Again, I don't see why using `fac` here again.  Now looking again, I think there is a mess with `fac`, `facmin`, `facmax` (the latter two are not used at all).

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+
+        #--#####################################################################
+        # Interpolation for dense output
+        #--#####################################################################
+        # interpolate specified points
+        while iter <= length(tspan) && tdir*tspan[iter]<tdir*t_np1
+            h_iter = tspan[iter]-t[n]
+
+            # Calculation of g_{j}(n)coefficients for j=0,...,k+1;
+            g_coefs!(g,c,h,t,tspan[iter],n,min(n,steporder+1))
```
mauro3: You'll have to explain again why these need recalculating.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+
+            # Calculation of  ϕstar_{j}(n), and ϕ_{j}(n) coef. for j=0,...,k;
+            ϕ_and_ϕstar_coefs!(b, ϕ, ϕstar,dy[n],t,tspan[iter],n,1:min(max(n-1,1),steporder+1))
+
+            yout_iter = y[n]
```
mauro3: You're re-bind the vector in `y[n]` to `yout_iter`. If below, you'd update `yout_iter` in-place (as you should) then `y[n]` is also updated!

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            for j= 1:steporder
+                yout_iter += h_iter*g[j]*ϕstar[2,j]
```
mauro3: in-place

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            end
+
+            #CE iteration for more precise interpolation
+            if n>=steporder+1
+                dy_iter=F(tspan[iter],yout_iter)
+                ϕ_coefs!(ϕ, ϕstar,dy_iter,1:steporder+1)
+                yout_iter = yout_iter + h_iter*g[steporder+1]*ϕ[3,steporder+1]
```
mauro3: in-place

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+                # Harier equation 7.4: If ||LE_{k+1}|| =< 1
+                # then step size is accepted and use next_h for next step
+                # Else, then step size is rejected, reattempt with fac*next_h
+                if err_kp1 <= 1
+                    successful_step = true
+                else
+                    successful_step = false
+                    h = fac*next_h # make sure fac < 1
+                end
+            end # End of error check for current step size h
```
mauro3: I think you ought to resize `h_next` to fit into tspan here, instead of https://github.com/JuliaLang/ODE.jl/pull/106/files#diff-43ccd017e404a50f115b60e06e7f1ac5R260

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            push!(tout, t_np1)
+            push!(yout, y_np1)
+        end
+
+        # restore values of coefficients for choosen step size
+        g_coefs!(g,c,h,t,t_np1,n,min(n,steporder+1))
+        ϕ_and_ϕstar_coefs!(b, ϕ, ϕstar,dy[n],t,t_np1,n,1:min(max(n-1,1),steporder+1))
+        ϕ_coefs!(ϕ, ϕstar,dy_np1,1:steporder+1)
+
+        # push work variables
```
mauro3: You'll have to push a copy here to make above in-place loops work.  Otherwise these will be modifed.

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+        #   See Harier pg 424 for indepeth description
+        #--#####################################################################
+
+        current_steporder = steporder
+        if n<=order+1
+            # Need to run the first few steps at reduced order for it requires
+            # the last 'order' number of points
+            steporder = n==1 ? 1 : n-1
+
+        elseif adaptive_order && steporder>1
```
mauro3: What is the `steporder>1` test for?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            # First, compute truncation errors for implicit orders: k-1,k
+            # against current k+1
+            yerr_km1 = h*(g[steporder] - g[steporder-1])*ϕ[3,steporder]
```
mauro3: in-place

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            err_km1 = stepsize_hw92!(h, tdir, y[n], y_np1, yerr_km1, steporder,
+                            timeout_VS,dof, abstol, reltol, maxstep, norm)[1]
+
+            yerr_k = h*(g[steporder+1] - g[steporder])*ϕ[3,steporder+1]
```
mauro3: in-place

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+            if max(err_km1,err_k) <= err_kp1
+                steporder = max(order,steporder - 1)
+
+            elseif consec_const_steps >= steporder && timeout_VO==0 && n>steporder + 3
+                # Given enough consecutive steps taken at a constant step size
+                # we compute truncation error for order implicit k+2
+                ϕ_and_ϕstar_coefs!(b, ϕ, ϕstar,dy[n],t,t_np1,n,steporder+2)
+                ϕ_coefs!(ϕ, ϕstar,dy_np1,steporder+3)
+
+                yerr_kp2 = h*γ_star(steporder + 3)*ϕ[3,steporder+3]
```
mauro3: in-place

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+        #--#####################################################################
+        # Prepare for next step
+        #--#####################################################################
+
+        # Shift data from this step, from the "current step" position to the
+        # "previous step" position for ϕstar and ϕ
+        # TODO: implement with DataStructures.CircularBuffer
+        ϕstar[1,1:min(max(n-1,1),current_steporder+1)] =
+                                ϕstar[2,1:min(max(n-1,1),current_steporder+1)]
+        ϕ[1,1:min(max(n-1,1),current_steporder+1)] =
```
mauro3: `ϕ[1,:]` is never used, or is it?

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+                                ϕ[2,1:min(max(n-1,1),current_steporder+1)]
+
+        # Update stepsize pointers
```
mauro3: There are no "pointers" here (that is a fairly specific computer science term).

src/adams_methods.jl
```diff
@@ -0,0 +1,655 @@
...
+    c = zeros(dim,dim)
+    g = zeros(dim)
+    b = zeros(dim)
+    @compat ϕstar = Array{T,2}(2,dim)
+    @compat ϕ = Array{T,2}(3,dim)
+    return c,g,b,ϕstar, ϕ
+end
+
+
+function make_consistent_types(fn, y0, tspan)
```
mauro3: It would be good to have this refactored with the existing one: just make the one taking a `btab` too call this one.  ((Almost) never copy-paste code).  Also move to ODE.jl-file.

mauro3:
Lot's of comments! It generally looks good, thanks!  Once you update to using more in-place operations, it would be good to see the benchmarks again.

src/adams_methods.jl
obiajulu: what do you think of an `@assert` for this? I implemented one.

src/adams_methods.jl
obiajulu: Great,  I added the note and the same `@assert` as above.

src/adams_methods.jl
obiajulu: Added the definition

src/adams_methods.jl
obiajulu: This better aligns with the convention, so yes, I think so. Made the change.

src/adams_methods.jl
obiajulu: Done and done.

src/ODE.jl
obiajulu: I changed to naming to `ode4am`, which fits with conventions better. And yeah, I think we were planning to open the PR against the original repo, and save the JuliaODE fork for work on iterative solvers.

src/adams_methods.jl
obiajulu: I added more docstrings. I figure it is best do keep formal documentation.

src/adams_methods.jl
obiajulu: Done.

src/adams_methods.jl
obiajulu: Sounds good, I'll remove these.

src/adams_methods.jl
obiajulu: Great, thanks. I just have now 

```
    yout = Array(Ty, 1)
    yout[1] = copy(y0)
```

src/adams_methods.jl
obiajulu: Great, I changed this.

src/adams_methods.jl
obiajulu: Great, thanks!

src/adams_methods.jl
obiajulu: For some reason I get a 

```
ERROR: LoadError: UndefVarError: t_np1 not defined
```
when I don't preallocate `t_np1`. I looked more into this, and I think it is a scope issue. Since I need `t_np1` for both all three parts of the time stepping loop (value update, step size validation, and order control) do you think it makes sense to declare the variable here?

src/adams_methods.jl
obiajulu: Yes indeed, it seems there is a bug somewhere else. I'm currently looking for it.

src/adams_methods.jl
obiajulu: I run into the same issues of scope as with `t_np1` when I delete the initial declaration.

src/adams_methods.jl
obiajulu: Duly noted and changed.

src/adams_methods.jl
obiajulu: `c` actually just holds real numbers.

src/adams_methods.jl
obiajulu: Never used, I'll refactor this.

src/adams_methods.jl
obiajulu: Sounds good.

src/adams_methods.jl
mauro3: I'd use an `if`

src/adams_methods.jl
mauro3: More compact `yout = Ty[copy(y0)]`, but no need to change it.

src/adams_methods.jl
mauro3: Yes, if a variable is only defined inside a loop, it will not be available outside.  But delete the comment.

src/adams_methods.jl
mauro3: Ok, then define `dy_np1` before the loop, it needn't be the right size but it needs to be of the right type.

src/adams_methods.jl
mauro3: By real you mean `Float64`?

If you were to calculate with BigFloats then `c` should also be BigFloats to avoid precision loss.

src/adams_methods.jl
obiajulu: Sounds good, I'll fix this right away.

src/adams_methods.jl
obiajulu: Oh, yeah, I see your point here, sorry. I'll fix this.

src/adams_methods.jl
obiajulu: Thanks.

src/adams_methods.jl
obiajulu: So this is exactly what I do. I am evaluating the integral of the polynomial from t_n to each different tspan point between t_n and t_{n+1}. This, however, requires calculating new g(n) coefficients. See the following formula from Harier vol 1, page 398
integration

src/adams_methods.jl
obiajulu: Sure thing, makes sense.

src/adams_methods.jl
obiajulu: I was trying to follow a protocol similar to what is outlined here. Though, I think you are right, and that the extra `fac*next_h` is unnecessary. [automatic-step-error-control.pdf](https://github.com/JuliaLang/ODE.jl/files/361791/automatic-step-error-control.pdf)

src/adams_methods.jl
obiajulu: I just removed `fac` for now. I'll see if it is necessary to add some safety "buffer" when I re-benchmark

src/adams_methods.jl
obiajulu: This was just a heuristic. I notice that I would get a little bit better performance on threebody and PLEI with `h/4` for the second step. I think, though, it might point to the fact that it is worthwhile looking into using a high order runge kutta method for the first few initial data points, instead of the gradual order build up, since such small changes in step sizes in the beginning do show a noticeable difference in performance.

src/adams_methods.jl
obiajulu: Great, made the change.

src/adams_methods.jl
obiajulu: Yes and no. Yes, this change would mean that there are no longer consecutive steps of the same size. However, seeing as this also means this is the last step, and `consec_const_steps` is used for changing the order for the next step, it won't matter if it is changed by this part of the code.

src/adams_methods.jl
obiajulu: Great, thanks.

src/adams_methods.jl
obiajulu: Yep, fixed it.

src/adams_methods.jl
obiajulu: So I think I found that the bug, at least with this specific inplace loop, was `copy!()`. Replacing this with 

```
            y_np1 = copy(y[n])
            for j= 1:steporder
                for i=1:length(y_np1)
                    y_np1[i] += dt*g[j]*ϕstar[2,j][i]
                end
            end
```
did the trick. I'm not sure if this may be a bug with `copy!()`.

src/adams_methods.jl
obiajulu: All good now

src/adams_methods.jl
obiajulu: Done

src/adams_methods.jl
obiajulu: Great, thanks!

src/adams_methods.jl
obiajulu: Great, thanks!

src/adams_methods.jl
obiajulu: Yup, made the change!

src/adams_methods.jl
obiajulu: I made these changes. I'm curious, though, why it is currently seeming not to cause any issues?

src/adams_methods.jl
obiajulu: Yup, thanks!

src/adams_methods.jl
obiajulu: Great, thanks.

src/adams_methods.jl
obiajulu: Great, thanks.

src/adams_methods.jl
obiajulu: Okay, I'll do this instead. Thanks.

src/adams_methods.jl
obiajulu: Okay, I'll instead just declare them here, for scope.

src/adams_methods.jl
obiajulu: Notice in line 482, we need the difference 
```
g[steporder] - g[steporder-1]
```
and so if `steporder` is one, then `g[steporder-1]` is out of range.

src/runge_kutta.jl
```diff
@@ -161,6 +161,7 @@ ode1(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_feuler)
 ode2_midpoint(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_midpoint)
 ode2_heun(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_heun)
 ode4(fn, y0, tspan) = oderk_fixed(fn, y0, tspan, bt_rk4)
+ode78f(fn, y0, tspan) = oderk_adapt(fn, y0, tspan, bt_feh78)
```
mauro3: This shouldn't be here, right?

src/runge_kutta.jl
```diff
@@ -289,6 +290,7 @@ function oderk_adapt{N,S}(fn, y0::AbstractVector, tspan, btab_::TableauRKExplici
     islaststep = abs(t+dt-tend)<=eps(tend) ? true : false
     timeout = 0 # for step-control
     iter = 2 # the index into tspan and ys
+
```
mauro3: Do not introduce a new-line here.

test/runtests.jl
```diff
@@ -12,12 +11,14 @@ solvers = [
            ODE.ode4,
            ODE.ode4ms,
            ODE.ode5ms,
+           ODE.ode_am,
```
mauro3: shouldn't this be `ODE.ode4am`?

src/adams_methods.jl
```diff
@@ -0,0 +1,637 @@
+"""
+    [T,X] = ode_ms(ODEFUN, Y0, TSPAN,ORDER=4) with TSPAN = [T0:dt:TFINAL]
```
mauro3: This is a bit matlab-y, at least change it to:
`t,y = ode_ms(ODEFUN, Y0, TSPAN, ORDER=4)`
but I would use lower-case.

src/adams_methods.jl
```diff
@@ -0,0 +1,637 @@
...
+
+Fixed-Equal Steps Adams-Bashforth multi-step solver for nonstiff problems
+
+(Main ref: Hairer & Wanner 1996, Vol I, p.359-360, 372)
+
+#Note: If method is run at order k, the first step is taken at order 1,
```
mauro3: remove `#`

src/adams_methods.jl
```diff
@@ -0,0 +1,637 @@
...
+#Note: this function works for both scalar-like and array-like input
+
+#Note: If method is run at order k, the first step is taken at order 1,
+and the order increases each step until it is order k on the kth step.
+Thus results may be of lower accuracy than expected.
+"""
+
+##Function: ode_am(F,y0, t, order(optional))
+##order is set to 4 by default
+function ode_am(F::Function,y0, t,order::Integer = 4)
```
mauro3: drop the `::Function` as types can be called too.

src/adams_methods.jl
```diff
@@ -0,0 +1,637 @@
...
+                                1//2      1//2     0         0
+                                -1//12    8//12    5//12     0
+                                1//24     -5//24   19//24    9//24]
+
+
+
+"""
+    [T,X] = ODE113(ODEFUN, Y0, TSPAN,ORDER=4;kwords...)
+
+with TSPAN = [T0:TFINAL] or[T0,T1,...,TFINAL] for dense output
```
mauro3: should read `[T0, TFINAL] or[T0,T1,...,TFINAL] for dense output`

src/adams_methods.jl
mauro3: But drop this particular doc-string.  As a rule only document one method per generic function, unless you have to document others.

src/adams_methods.jl
mauro3: Tnx

src/adams_methods.jl
```diff
@@ -0,0 +1,637 @@
...
+
+    # TODO: use a better data structure here (should be an order-element circ buffer)
+    dy = similar(y)
+    for i = 1:length(tspan)-1
+        # Need to run the first several steps at reduced order
+        steporder = min(i, order)
+        dy[i] = F(tspan[i], y[i])
+
+        # Check step size is equal or very near equal
+        if abs((dt[i] - dt[1])/dt[i]) > .01
```
mauro3: `!isapprox(dt[i], dt[1])`

test/runtests.jl
obiajulu: Yes, thanks for the catch. `ode_am` defaults to `ode4am`, but it is best to be more precise.

src/adams_methods.jl
obiajulu: Sure thing, I'll make the change. Thanks.

src/adams_methods.jl
obiajulu: Thanks

src/runge_kutta.jl
obiajulu: Thanks.

src/adams_methods.jl
obiajulu: Thanks.

src/adams_methods.jl
obiajulu: Yes, you're right.

src/adams_methods.jl
obiajulu: Great, I didn't know about this function, I'll use it instead.

src/runge_kutta.jl
obiajulu: Yes, you are correct. I used it for testing purposes. I'll remove it.

src/adams_methods.jl
```diff
@@ -0,0 +1,636 @@
...
+                                            abstol, reltol, maxstep, norm)
+
+                # Harier equation 7.4: If ||LE_{k+1}|| =< 1
+                # then step size is accepted and use next_dt for next step
+                # Else, then step size is rejected, reattempt with fac*next_dt
+                if err_kp1 <= 1
+                    successful_step = true
+                else
+                    successful_step = false
+                    dt = .75*next_dt # make sure fac < 1
```
mauro3: Should not `fac` be used here?  Also, do we need `fac` as an keyword option to ode113?

src/adams_methods.jl
```diff
@@ -0,0 +1,636 @@
...
+                end
+            end
+
+            push!(tout,tspan[iter])
+            push!(yout, copy(yout_iter))
+            iter += 1
+        end
+
+        if iter <= length(tspan) && tspan[iter] == t_np1
+            # handling when tspan and choose time steps overlap
```
mauro3: "choosen" typo

src/adams_methods.jl
```diff
@@ -0,0 +1,636 @@
...
+
+                # Calculate ||LE_{k+1}|| and next stepsize
+                for i = 1: dof
+                    yerr_kp1[i] = dt*(g[steporder+2] - g[steporder+1])*ϕ[3,steporder+2][i]
+                end
+                err_kp1, next_dt, timeout_VS = stepsize_hw92!(dt,tdir,y[n],y_np1,
+                                            yerr_kp1, steporder, timeout_VS,dof,
+                                            abstol, reltol, maxstep, norm)
+
+                # Harier equation 7.4: If ||LE_{k+1}|| =< 1
```
mauro3: "Hairer"

src/adams_methods.jl
```diff
@@ -0,0 +1,636 @@
...
+                                ϕstar[2,1:min(max(n-1,1),current_steporder+1)]
+
+        # Update stepsize variables
+        prev_dt = dt
+        dt = next_dt
+        # Increment our step
+        n= n+1
+    end #while loop over steps
+
+    if points == :steps
```
mauro3: There is no option `:steps`.  `:specified`?

src/adams_methods.jl
```diff
@@ -0,0 +1,636 @@
...
+            push!(tout, copy(t_np1))
+            push!(yout, copy(y_np1))
+        end
+
+        # restore values of coefficients for choosen step size
+        g_coefs!(g,c,dt,t,t_np1,n,min(n,steporder+1))
+        ϕ_and_ϕstar_coefs!(b, ϕ, ϕstar,dy[n],t,t_np1,n,1:min(max(n-1,1),steporder+1))
+        ϕ_coefs!(ϕ, ϕstar,dy_np1,1:steporder+1)
+
+        # push work variables
```
mauro3: This has nothing to do with dense output, maybe move to above dense output loop?

src/adams_methods.jl
```diff
@@ -0,0 +1,636 @@
...
+
+            # Calculation of  ϕstar_{j}(n), and ϕ_{j}(n) coef. for j=0,...,k;
+            ϕ_and_ϕstar_coefs!(b, ϕ, ϕstar,dy[n],t,t_np1,n,1:min(max(n-1,1),steporder+1))
+            #Note that  1<= min(max(n-1,1),steporder+1) <= n-1 and steporder+1
+
+            # Calculate next value, y_np1, with order = k = steporder + 1. See
+            # equation (5.5) in Hairer et al, Vol 1, where we note that here
+            # y_np1 = y_{k}(t_{n+1})
+
+            y_np1 = copy(y[n])
```
mauro3: `copy!(y_np1, y[n])` to avoid allocations.

src/adams_methods.jl
```diff
@@ -0,0 +1,636 @@
...
+            else
+                # We calculate local truncation error, LE_{k+1}; the difference
+                # between implicit results for order k and k+1.
+
+                # We need ϕ_{k+1}(n+1) and g_{k+1}(n);the latter has already
+                # been calculated above, the former still needs calculation
+                ϕ_coefs!(ϕ, ϕstar,dy_np1,steporder+2)
+
+                # Calculate ||LE_{k+1}|| and next stepsize
+                for i = 1: dof
```
mauro3: This `yerr_kp1` gets overwritten in `stepsize_hw92!`.  If this loop is supposed to do something, then it doesn't do it, otherwise delete.  In the other places where `stepsize_hw92!` gets called, the same occurs.

src/adams_methods.jl
```diff
@@ -248,7 +248,16 @@ function ode113(F, y0::AbstractVector, tspan, order::Integer=4;
 
     # initialization of arrays used in estimation y_np1 given dy[n],y[n], dt and
     # list order timestep are  initialized to a maxorder x maxorder dimension
-    c,g,b,ϕstar, ϕ = intialize_coef_arrays(max_order+3,Tdy,Et)
+    c,g,b,ϕstar, ϕ = intialize_coef_arrays(max_order+3,Tdy,Et,y0)
+
+    if max_order <= 13
+        γ_star = Float64[1, -0.5,-0.0833333,-0.0416667,-0.0263889,
```
obiajulu: Move out \gamma_star

obiajulu: Make const, BigFloat and move out.

src/adams_methods.jl
obiajulu: Yes, I made the switch, and, the choice of `fac` does change the results slightly, so it is a question of if we want to give the user this the ability to easily change this. However, to fit conventions, I'll remove the keyword option.

src/adams_methods.jl
obiajulu: Thanks!

src/adams_methods.jl
obiajulu: Thanks, I made this typo a few times. They have all been corrected now.

src/adams_methods.jl
obiajulu: Yeah, I wanted to ask you about this. So, currently the options for `points` are 

- `specified`
- `all`, which is the union of points `specified` and points which were used to step through the time range

I thought, it might be worthwhile to also have a `steps` options which is `all` - `specified`; namely, the points actually used to calculate the solution. Though, it is mostly  because it is very easy to add this extra option.

obiajulu:
After these rounds of revisions, these are the current performance benchmarks, for ` abstols = 10.0.^(-5:-1:-15))`. A slight time decrease is observed, however, still not to the degree anticipated after the removal of dependency on `Polynomials`. However, the reason I think this is so is that, even though this dependency took up a significant part of the ProfileView plot, these function calls may not have been very expense. This is further confirmed by an experiment I performed: taking the first commit of the PR, and using the gamma_star array look up there, yielded non-stellar time decreases. 
threebody_2

plei_2

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+
+
+#Based on Hairer equation 5.9. Just calculates the ϕ coefficients
+"compute ϕ_{n+1,j} coefficients for ode113"
+function ϕ_coefs!(ϕ, ϕstar, dy_n, range_of_indices)
+    for j in range_of_indices
+      if j == 1     #0
+        ϕ[3, j] = copy(dy_n)
+      else j > 1    #0
+        #Note: since n>=j, when this runs, n>=2 (so n-1>=1) and
```
mauro3: Clean up comment.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+      end
+    end
+    return nothing
+end
+
+################################################################################
+# Helper functions initializing/resizing arrays
+################################################################################
+"used to initialize working arrays used in ode1113"
+function intialize_coef_arrays(dim, T, Et, y0)
```
mauro3: `y0` is not used.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+    end
+    return γ_star
+end
+################################################################################
+# Helper functions for calculating coefficients from recurence relationships
+################################################################################
+#Based on Hairer Lemma 5.1 on page 399
+#Calculates c_j,q and g_n,j for n and j =1:steporder
+"compute c_{j,q} and g_{n,j} coefficients for ode113"
+function g_coefs!(g,c,dt,t,t_np1,n,steporder)
```
mauro3: the argument `steporder` is always called with `steporder+1` or `+2`.  So is this really the steporder?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+    for j = 1:steporder
+      for q = 1:(steporder) - (j-1)
```
mauro3: Use parenthesis for clarity: `for q = 1:(steporder - (j-1))`.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+        # initial guess at a step size
+        # dt, tdir, dy0 = hinit(F, y0, tstart, tfinal, 3, reltol, abstol)
+        dt, tdir, dy0 = hinit(F, y0, tstart, tfinal, 1, reltol, abstol)
+
+    else
+        tdir = sign(tfinal - tstart)
+    end
+
+    n = 1     #`n` counts the time steps. We start at n=1, the first step
+    steporder=1 # `steporder` is the order of the AB method at the current step
```
mauro3: I think this comment could be clearer: `k+1` is the order of the implicit method (see e.g. p.361).  H&W run their sums/products over `j=0:k-1` or `j=0:k`, whereas here `j=1:k` or `j=1:k+1` (so `j` can be used as index).  So, it seems `steporder` is the order of the explicit method and `steporder+1` of the implicit, right?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+            #---################################################################
+
+            # Calculation of g_{j}(n)coefficients for j=0,...,k+1;
+            g_coefs!(g, c, dt, t, t_np1, n, min(n, steporder+2))
+
+            # Calculation of  ϕstar_{j}(n), and ϕ_{j}(n) coef. for j=0,...,k;
+            ϕ_and_ϕstar_coefs!(b, ϕ, ϕstar, dy[n], t, t_np1, n, 1:min(max(n-1, 1), steporder+1))
+            #Note that  1<= min(max(n-1,1),steporder+1) <= n-1 and steporder+1
+
+            # Calculate next value, y_np1, with order = k = steporder + 1. See
```
mauro3: Clearer would be "Calculate predictor with explicit method of order `steporder`"

mauro3:
Could you post your IVPTestSuite script as a gist?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+            #---################################################################
+            # Correction-Evaluaion (CE) Loop for implicit solver
+            #     See Hairer p. 360 for a general description of this method
+            #---################################################################
+
+            # Set CE_loops to be how many correction-evaulation cycles
+            # desired. As least one loop is necessary, for the sake of the
+            # adaptive step selection
+            CE_loops = 1
+            if n>=steporder+1
```
mauro3: Shouldn't the trapezoidal rule be used on the first step?  But this only runs on the second step, right?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+                    successful_step = false
+                    dt = fac * next_dt
+                end
+            else
+                # We calculate local truncation error, LE_{k+1}; the difference
+                # between implicit results for order k and k+1.
+
+                # We need ϕ_{k+1}(n+1) and g_{k+1}(n);the latter has already
+                # been calculated above, the former still needs calculation
+                ϕ_coefs!(ϕ, ϕstar, dy_np1, steporder+2)
```
mauro3: Is it correct to call this with a non-range as last argument?  Aren't the `ϕ[3, j]` dependent on `ϕ[3, j-1]`, so all would need to be calculated.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+        if tdir*(t_np1 + next_dt) >= tdir*(tfinal)
+            next_dt = tfinal-t_np1
+        end
+
+        #--#####################################################################
+        # Interpolation for dense output
+        #--#####################################################################
+        # interpolate specified points
+        while iter <= length(tspan) && tdir*tspan[iter] < tdir*t_np1
+            h_iter = tspan[iter]-t[n]
```
mauro3: This should be `t_iter`.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+        ϕ_coefs!(ϕ, ϕstar, dy_np1, 1:steporder+1)
+
+        # push work variables
+        push!(y,  copy(y_np1))
+        push!(dy, copy(dy_np1))
+        push!(t,  copy(t_np1))
+
+        #--#####################################################################
+        # Variable order selection for next step
+        #   Note run until a success step was taken for n-th step
```
mauro3: "Not run..."

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+
+        #--#####################################################################
+        # Prepare for next step
+        #--#####################################################################
+
+        # Shift data from this step, from the "current step" position to the
+        # "previous step" position for ϕstar and ϕ
+        # Note: ϕ[2, :] and ϕ_star[2, :] corresponds to ϕ_n and ϕ*_n, and
+        # ϕ_star[1, :] to ϕ*_{n-1}, etc
+        # TODO: implement with DataStructures.CircularBuffer
```
mauro3: I don't think this needs a circular buffer as only two states are needed.  Remove comment.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+#n and j  in range_of_indices
+"compute ϕ_{n,j} and ϕ*_{n,j} coefficients for ode113"
+function ϕ_and_ϕstar_coefs!(b, ϕ, ϕstar, dy_n, t, t_np1, n, range_of_indices)
+    for j in range_of_indices #0:k-1
+      if j == 1     #0
+        b[j] = 1
+        dy_c = copy(dy_n)
+        ϕ[2, j] = dy_c
+        ϕstar[2, j] = dy_c
+      else j > 1    #0
```
mauro3: This `j > 1` doesn't do anything, remove.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+end
+
+
+#Based on Hairer equation 5.9. Just calculates the ϕ coefficients
+"compute ϕ_{n+1,j} coefficients for ode113"
+function ϕ_coefs!(ϕ, ϕstar, dy_n, range_of_indices)
+    for j in range_of_indices
+      if j == 1     #0
+        ϕ[3, j] = copy(dy_n)
+      else j > 1    #0
```
mauro3: Ditto

src/adams_methods.jl
obiajulu: I just removed it because I no longer think it is all too necessary. Thanks.

src/adams_methods.jl
obiajulu: Thanks, made the changes.

src/adams_methods.jl
obiajulu: Great, thanks

src/adams_methods.jl
obiajulu: Yes, this is true. It was indeed a large source of off-by-one errors. Using `stageorder` in replace of `steporder` does seem like a good approach to reduce the potential confusion.

src/adams_methods.jl
obiajulu: Yes, I think that will more easily explain what is going on in a first read through of the code.

src/adams_methods.jl
obiajulu: Yes, I think this was an hangover from a previous version which had an off-by-one indexing error.

src/adams_methods.jl
obiajulu: Thanks!

src/adams_methods.jl
obiajulu: Thanks

src/adams_methods.jl
obiajulu: Great, thanks

src/adams_methods.jl
obiajulu: Great, thanks for the catch

src/adams_methods.jl
obiajulu: Thanks again

src/adams_methods.jl
```diff
@@ -0,0 +1,628 @@
...
+      g[j]=c[j,1]
+    end
+    return nothing
+end
+
+
+#Based on Hairer equation 5.9. Calculates b, ϕ, ϕstar for
+#n and j  in range_of_indices
+"compute ϕ_{n,j} and ϕ*_{n,j} coefficients for ode113"
+function ϕ_and_ϕstar_coefs!(b, ϕstar_nm1, ϕstar_n, ϕ_n,dy_n,t,t_np1,n,range_of_indices)
```
mauro3: put arguments which are modified in-place first.  Document which are modified in-place.

src/adams_methods.jl
```diff
@@ -0,0 +1,628 @@
...
+        ϕstar_n[j] = b[j]*ϕ_n[j]
+      end
+    end
+    return nothing
+end
+
+
+#Based on Hairer equation 5.9. Just calculates the ϕ coefficients
+"compute ϕ_{n+1,j} coefficients for ode113"
+function ϕ_coefs!(ϕstar_n,ϕ_np1,dy_n,range_of_indices)
```
mauro3: As above.

obiajulu:
After another batch of changes, I think that this PR needs none or little further work until review by others in the ODE.jl community.  We once again run benchmarks on the performance of `ode113` against `ode45` and `ode78` for `abstols = 10.0.^(-5:-1:-14))`

figure_13
figure_14

The following script can be run to reproduce these results:

```
Pkg.checkout("IVPTestSuite","terminalmode")
using IVPTestSuite.QuickSuites
runtestsuite(ODEsolverfns = [ODE.ode113,ODE.ode45, ODE.ode78],sundialsolverfns= [],DASSLsolverfns = [], abstols = 10.0.^(-5:-1:-14))
plottestsuite()
```

or through use of the notebook mode of IVPTestSuite (link pending)

pwl:
These plots could use a log scale for Walltime.

mauro3:
Good point @pwl!

ChrisRackauckas:
Note that this really isn't the test you want to be doing for Adams methods. You should test it on equations which have a large function calculation cost. When function calculations are cheap, RK methods will pretty much always win (see Hairer, or papers by Shampine or Verner or etc. there's a large list out there). A good test case would be something like a MOL discretization where you have a linear system to solve at each step, or something that has a root finding / optimization call inside of the function.

ChrisRackauckas:
Also, for reference, did you do Nystrom or variable coefficient implementations?

obiajulu:
@ChrisRackauckas, thanks, this is a really good point. I'll look into getting a test case with a higher function evaluation cost. Right now, these two test cases (PLEI and threebody) are the only nonstiff test cases in IVPTestSuite, so it would be good to get some more in there regardless. Could you post a link to more information on the MOL discretization and/or other test cases with high function evaluations?

Also, I used the variable coefficient implementation from Hairer et al.

ChrisRackauckas:
For MOL discretization, just take a semilinear Heat equation with and make a reaction term with bounded but hard to compute coefficients (like the ROBER model, except with space to give a larger vector, and bound the coefficients so it's not too stiff). That should easily give a test case where function evaluations dominate the cost and an Adams method could be efficient. I may code up some example problem within the next week if you're willing to wait.

It would be interesting to test your implementation vs LSODE. ODEInterface will probably have it wrapped soon, so when that's together I'll run some benchmarks between them in DifferentialEquations.jl and let you know how that goes.

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.3
 Polynomials
 Compat 0.4.1
+ProgressMeter
```
ChrisRackauckas: This is very odd to introduce into this PR and introduces a new (albeit small) dependency. Shouldn't this be own own PR?

mauro3: Yes, Chris is right.  One PR one topic.  So PrgressMeter belongs to a different PR.

obiajulu: Yes, right you all are. I was mostly experimenting here and forgot to start a new branch. I'm going to revert these changes and open a different PR elsewhere.

obiajulu:
Another round of improvements to `ode113`, which should bring the solver to complete correctness. There was a lingering off-by-one error that I finally got to the bottom of and fixed. Performance is comparable to previous iterations, as confirmed by results below:

figure_3
figure_4

mauro3:
Yay, faster than ode78!

ChrisRackauckas:
This is still labelled WIP: is there anything keeping it from being merged?

mauro3:
Last I talked with @obiajulu (September), was that he should request a review from someone else.  Then we'd merge.  Would you be willing to look this over?  I think at this stage we're just looking for correctness-issues.  Any performance tuning (apart from easy stuff) would be later/never.

ChrisRackauckas:
Sure, I'll see if I can review this before the middle of the week.

mauro3:
Thanks! Note though, I will not have time to look at this before next year. So no hurry on my part.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+"""
+
+##Function: ode_am(F,y0, t, order(optional))
+##order is set to 4 by default
+function ode_am(F,y0, t,order::Integer = 4)
+    if (1 <= order <= 4)
+        b_imp = am_imp_coefficients3
+        b_exp = ms_coefficients4
+    else
+        #calculating higher order coefficients for implicit Adams Moulton method
```
ChrisRackauckas: Should this part for determining more coefficients instead be a separate function?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+
+    # initialization of time span, which will have variable length
+    tspan = convert(Vector{Et}, tspan)
+    tstart = tspan[1]
+    tfinal = tspan[end]
+    t = [tstart]
+
+    # initialzation of y[n], which will have variable length as well
+    # TODO: fix after https://github.com/JuliaLang/julia/issues/16667
+    #y_n = deepcopy(y0)
```
ChrisRackauckas: To be fully compatible with `AbstractVector` this needs to deepcopy. You can however

```julia
if typeof(y0) <: Vector
  y_n = copy(y0)
else
  y_n = deepcopy(y0)
end
```

to be slightly more performant in the standard case that it's a standard vector.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+    #initialization of step size dt, which will have variable length as well
+    dt = initstep # current trial stepsize for a give step
+    prev_dt = dt # the stepsize of the last successful step
+    next_dt = 0.0 # stepsize to be used in next step
+
+    if dt == 0.
+        # initial guess at a step size
+        # dt, tdir, dy0 = hinit(F, y0, tstart, tfinal, 3, reltol, abstol)
+        dt, tdir, dy0 = hinit(F, y0, tstart, tfinal, 1, reltol, abstol)
+        dt = dt
```
ChrisRackauckas: ?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+Fixed-Equal Steps Adams-Bashforth multi-step solver for nonstiff problems
+
+(Main ref: Hairer & Wanner 1996, Vol I, p.359-360, 372)
+
+Note: If method is run at order k, the first step is taken at order 1,
+and the order increases each step until it is order k on the kth step.
+Thus results may be of lower accuracy than expected.
+"""
+
+function ode_ms(F, y0, tspan, order::Integer)
```
ChrisRackauckas: These should have a trailing `;kwargs...` to match the fixes of https://github.com/JuliaDiffEq/ODE.jl/pull/119

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+        y[i+1] = y[i]
+        for j = 1:steporder
+            y[i+1] += dt[i]*b[steporder, j]*dy[i-(steporder-1) + (j-1)]
+        end
+    end
+    return vcat(tspan), y
+end
+
+# Use order 4 by default
+ode4ms(F, y0, tspan) = ODE.ode_ms(F, y0, tspan, 4)
```
ChrisRackauckas: `;kwargs...`  here too

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+            y[i+1] += dt[i]*b_imp[steporder, j]*dy[i - (steporder -1) + j]
+        end
+
+        ##(E)valulate the function anew at corrected approximatiion t[i+1],y[i+1]
+        dy[i+1] = F(t[i+1], y[i+1])
+    end
+    return t, y
+end
+
+ode4am(F::Function,y0, t)= ode_am(F::Function,y0, t,4)
```
ChrisRackauckas: `;kwargs...`

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+            # Error Check for step size `dt` and Variable next step selection:
+            #     Check error to determine if step size should be accepted:
+            #     ->if yes, compute `next_dt` for next step
+            #     ->if no, use this `next_dt`` for next attempt at this step
+            #---################################################################
+
+            #Don't allow increases of steps until n >= num_stages + 2
+            if n==1
+                successful_step = true
+                next_dt = dt/4
```
ChrisRackauckas: what's this `dt/4` for?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+            #---################################################################
+            # Correction-Evaluaion (CE) Loop for implicit solver
+            #     See Hairer p. 360 for a general description of this method
+            #---################################################################
+
+            # Set CE_loops to be how many correction-evaulation cycles
+            # desired. As least one loop is necessary, for the sake of the
+            # adaptive step selection
+            CE_loops = 1
+            for loop_counter=1:CE_loops
```
ChrisRackauckas: This is fine for now, but it should really have the option to loop until some tolerance between the previous and the current is satisfied.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+                else
+                    successful_step = false
+                    dt = fac*next_dt
+                end
+            end # End of error check for current step size dt
+        end # End of loop for finding successfull step size for nth step
+            # repeats if success_step == false
+
+        #Keep step size constant if `dt` is close to `prev_dt`
+        if abs(next_dt/prev_dt)<2 && abs(next_dt/prev_dt)>=1
```
ChrisRackauckas: Maybe expose this "close" as an keyword option?

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+
+            push!(tout,tspan[iter])
+            push!(yout, copy(yout_iter))
+            iter += 1
+        end
+
+        if iter <= length(tspan) && tspan[iter] == t_np1
+            # handling when tspan and choosen time steps overlap
+            push!(tout, copy(t_np1))
+            push!(yout, copy(y_np1))
```
ChrisRackauckas: again with the `copy` vs `deepcopy`. Either restrict the function for `Vector` (or `Array`, have you tried it on `Array`? It might already work), or have a path for `deepcopy`.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+            iter += 1
+        elseif points == :all
+            push!(tout, copy(t_np1))
+            push!(yout, copy(y_np1))
```
ChrisRackauckas: `deepcopy` issue here again.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+            # CE iteration for more precise interpolation
+            if n>=num_stages+1
+                dy_iter=F(tspan[iter],yout_iter)
+                ϕ_coefs!(ϕ_np1, ϕstar_n,dy_iter,1:(num_stages)+1)
+
+                yout_iter += dt_iter*g[(num_stages)+1]*ϕ_np1[(num_stages)+1]
+            end
+
+            push!(tout,tspan[iter])
+            push!(yout, copy(yout_iter))
```
ChrisRackauckas: `deepcopy` issue here again.

src/adams_methods.jl
```diff
@@ -0,0 +1,626 @@
...
+
+        # Increment our step
+        n= n+1
+        if progressmeter
+          prog.counter = max(1,round(Int,1000.0*tdir*(t[n]-tstart)))
+          ProgressMeter.updateProgress!(prog; showvalues = [(:steps,n),(:current_t,t[n]),(:tfinal,tfinal)])
+        end
+    end #while loop over steps
+
+    return vcat(tout), yout
```
ChrisRackauckas: why the `vcat`?

mauro3:
I'll close this as it is unlikely to be resurrected.

ChrisRackauckas:
It was resurrected in spirit, as the core coefficient calculation algorithms were the starter code for those in the OrdinaryDiffEq.jl implementation: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/adams_utils.jl so thanks!

PR #107, opened 2016-07-11
example of a user defined type

jkozdon:


jkozdon:
There doesn't seem to be an example showing what's required for the user to define their own type.

examples/User_Type.jl
```diff
@@ -0,0 +1,43 @@
...
+
+typealias DataFloat Float64
+
+type Vec2
+  a::DataFloat
+  b::DataFloat
+end
+
+# Necessary for the solver
+Base.(:/)(x::Vec2, y::Real) = Vec2(x.a/y, x.b/y)
```
mauro3: To make this 0.5-proof: `@compat Base.:+`

examples/User_Type.jl
```diff
@@ -0,0 +1,43 @@
...
+Base.(:*)(y::Real, x::Vec2) = Vec2(y * x.a, y * x.b)
+Base.(:*)(x::Vec2, y::Real) = y*x
+Base.(:.*)(y::Real, x::Vec2) = y*x
+Base.(:+)(x::Vec2, y::Vec2) = Vec2(x.a + y.a, x.b + y.b)
+Base.(:-)(x::Vec2, y::Vec2) = Vec2(x.a - y.a, x.b - y.b)
+Base.norm(x::Vec2) = sqrt(x.a^2 + x.b^2)
+Base.zero(x::Type{Vec2}) = Vec2(zero(DataFloat), zero(DataFloat))
+Base.isnan(x::Vec2) = isnan(x.a) || isnan(x.b)
```
mauro3: This should be `ODE.isoutofdomain`

examples/User_Type.jl
```diff
@@ -0,0 +1,43 @@
...
+
+function f(t, y)
+  (a, b) = (y.a, y.b)
+  Vec2(-b, a)
+end
+
+let
```
mauro3: I'd leave the `let` block away.

examples/User_Type.jl
```diff
@@ -0,0 +1,43 @@
...
+  # Initial condtions
+  const start = Vec2(0.0, 0.1)
```
mauro3: These don't need to be `const` for performance as they are passed into the function as arguments.

mauro3:
Thanks, this is good!  Maybe a few sentences comments would be good: what the example is about, maybe a word about the ODE.

There is also the test-case `test/interface-tests.jl`.  You could mention it in your example.

mauro3:
Looks good to me.  Someone please merge this, unless there are objections.

ChrisRackauckas:
Looks good to me as well. Merging.

mauro3:
Ah, yes, we can merge now! ;-)  Thanks!

PR #108, opened 2016-07-12
Add some new examples

YingboMa:
I have added some new examples for ODE.jl package.


examples/README.md
```diff
@@ -8,3 +8,5 @@ You can also directly see the result without installing anything. Just follow th
 * [Simple differential equation](http://nbviewer.jupyter.org/github/JuliaLang/ODE.jl/blob/master/examples/Simple_Differential_Equation.ipynb)
 * [Lorenz Attractor](http://nbviewer.jupyter.org/github/JuliaLang/ODE.jl/blob/master/examples/Lorenz_Attractor.ipynb)
 * [Terminal Velocity](http://nbviewer.jupyter.org/github/JuliaLang/ODE.jl/blob/master/examples/Terminal_Velocity.ipynb)
+* [Duffing oscillator](https://github.com/YingboMa/ODE.jl/tree/ym/examples/examples)
```
mauro3: These need updating.

mauro3:
Thanks!

Here are the two examples rendered:
http://nbviewer.jupyter.org/github/YingboMa/ODE.jl/blob/57689392f2a03d713b9d3f38cdef37855c816457/examples/Duffing_oscillator.ipynb
http://nbviewer.jupyter.org/github/YingboMa/ODE.jl/blob/57689392f2a03d713b9d3f38cdef37855c816457/examples/Van_der_Pol_oscillator.ipynb

obiajulu:
Thanks @YingboMa, these look good. There seems, though, to be some text formatting issues on Van der Pol oscillator. It seems some text that ought not be bold is in fact boldfaced.

mauro3:
Travis build failure unrelated.

VdP: 

- cell 0: write `\sin` in the equations
- cell 3: make those `const`, otherwise performance suffers.
- the warmup is in cell 5 is not needed if you don't time the execution, drop it.
- I only realized now that you're doing the *forced* VdP.  Maybe show also a chaotic orbit?

Duff:
- cell 3: make those `const`
- mention somewhere that Duff is not stiff and that thus ode78 is expected to be better
- quote all variables, etc with backticks.
- I think this needs a bit more work.  It is not quite clear what you want to show.  Add some more text to develop it better.  In particular the error analysis.
- cell 0: use `\cos` in eq.
-

ChrisRackauckas:
@mauro3 's list looks pretty solid.

ChrisRackauckas:
How are the changes coming along?

PR #113, opened 2016-09-01
Enable Testing for Windows

ChrisRackauckas:
This would enable AppVeyor testing for Windows. These tests passed without any changes. Is anyone against this? Is there a reason it wasn't enabled before?


acroy:
It was enabled, but got disabled when Sundials.jl moved to JuliaMath ... So definitely +1 from my side (maybe you want to squash the commits before merging).

ChrisRackauckas:
Yes, the commits will be squashed for sure. No one needs to know I missed spelled AppVeyor at first and couldn't find out why the tests wouldn't run :+1: .

acroy:
Ups, sorry I thought this was for Sundials ... I think we never tested on Windows because there is no OS specific code involved and therefore there is no reason to expect that it fails there but works on Travis (which already covers two OSs).

ChrisRackauckas:
I see. I think it's important to make sure there's no OS specific code though. One thing AppVeyor has found for me in the past was accidentally using different kinds of Ints on 32 vs 64-bit, causing tests to fail (and sometimes convergence tests to fail) on 32-bit Windows.

ChrisRackauckas:
Merging this for consistency. Everything else in JuliaDiffEq is being setup to test with Windows, it doesn't hurt the CI times, and we will really want to know that Windows is actually supported (it tends to get ignored by developers since most users use Windows!).

PR #114, opened 2016-09-22
Update the oldest release testing to v0.4

ChrisRackauckas:
Now it should be v0.4, v0.5, and v0.6 instead of v0.3 v0.5 v0.6. We are already not compatible for v0.3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ os:
     - osx
     - linux
 julia:
-    - 0.3
+    - 0.4
     - release
     - nightly
 git:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,3 @@
-julia 0.3
+julia 0.4
 Polynomials
-Compat 0.4.1
+Compat 0.9


ChrisRackauckas:
Also updated the minimum version requirements to v0.4 since v0.3 tests were already failing. Compat minimum version was upped to v0.9.0. This is sensible sense it passes for v0.3+, and I'm sure if the Compat version is far too old it'll cause problems.

tkelman:
What are you using from 0.9.0? Minimum versions do NOT mean "the version I tested with." They mean "the lowest version that implements the features I'm using."

ChrisRackauckas:
I don't know when everything got added to Compat. Is there a simple chart for me to follow? When I need to add minimum versions for something like that, if I can't figure out what the version needs to be (mostly it's just Compat because I don't follow all of Compat's activity), then I use a version that I know it has been tested on because that's the earliest version that I can guarantee as working.

tkelman:
Usually cross-referencing https://github.com/JuliaLang/Compat.jl/blame/master/README.md to the feature you're using it pretty accurate. Sometimes you might need to go back earlier if someone had a reformatting-only commit to the readme.

ChrisRackauckas:
Thanks for the tip. I'll use that in the future.

PR #117, opened 2016-11-12
Reduce allocations in the Rosenbrock methods

ChrisRackauckas:
This changes uses of `eye(J) - ....` to `I - ...`. `I`, the `UniformScaling` operator, uses dispatch to perform this calculation without actually allocating the identity matrix, improving the speed in almost every case. 

Also, in `oderosenbrock` someone had `eye(J)/gamma/hs` ... that's the wrong equation. I fixed it to I/(gamma*hs). That's probably why there was a `#Fix Me!` right above it.

### DIFF
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -281,14 +281,14 @@ function ode23s(F, y0, tspan; reltol = 1.0e-5, abstol = 1.0e-8,
         end
 
         if size(J,1) == 1
-            W = one(J) - h*d*J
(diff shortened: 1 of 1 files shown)


dpsanders:
`1 / a / b` and `1 / (a*b)` are equal, since operators with the same precedence are parsed left to right.
However, the second version is clearly less (mentally) ambiguous (as you have just shown).

ChrisRackauckas:
Good to know.

PR #119, opened 2016-12-12
add common interface bindings

ChrisRackauckas:
This adds common interface bindings to ODE.jl. Currently it only includes the algorithms in the README:

- ode23Alg
- ode45Alg
- ode78Alg
- ode23sAlg

I had to put the `Alg` in the type name because otherwise it matches the function's name. If anyone has a better idea of how to handle this let me know.

### DIFF
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -22,7 +22,7 @@ export ode23s
 export ode4s
 
 # Common Interface
-export ODEJLAlgorithm
+export ODEjlAlgorithm
 
 ## complete function export list: see runtests.jl
 
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -420,7 +420,7 @@ const s4_coefficients = (0.5,
 ode4s_s(F, x0, tspan; jacobian=nothing) = oderosenbrock(F, x0, tspan, s4_coefficients...; jacobian=jacobian)
 
 # Use Shampine coefficients by default (matching Numerical Recipes)
-const ode4s = ode4s_s
+ode4s(F, x0, tspan; jacobian=nothing) = ode4s_s(F, x0, tspan; jacobian=nothing)
 
 const ms_coefficients4 = [ 1      0      0     0
                           -1/2    3/2    0     0
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -9,6 +9,8 @@ using DiffEqBase
 
 import DiffEqBase: solve
 
+include("algorithm_types.jl")
+
 ## minimal function export list
 # adaptive non-stiff:
 export ode23, ode45, ode78
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -21,6 +21,9 @@ export ode23s
 # non-adaptive stiff:
 export ode4s
 
+# Common Interface
+export ODEJLAlgorithm
+
 ## complete function export list: see runtests.jl
 
 ###############################################################################
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -157,7 +157,7 @@ include("runge_kutta.jl")
 
 # ODE_MS Fixed-step, fixed-order multi-step numerical method
 #   with Adams-Bashforth-Moulton coefficients
-function ode_ms(F, x0, tspan, order::Integer)
+function ode_ms(F, x0, tspan, order::Integer; kwargs...)
     h = diff(tspan)
     x = Array(typeof(x0), length(tspan))
     x[1] = x0
@@ -194,8 +194,8 @@ function ode_ms(F, x0, tspan, order::Integer)
 end
 
 # Use order 4 by default
-ode4ms(F, x0, tspan) = ode_ms(F, x0, tspan, 4)
-ode5ms(F, x0, tspan) = ODE.ode_ms(F, x0, tspan, 5)
+ode4ms(F, x0, tspan; kwargs...) = ode_ms(F, x0, tspan, 4; kwargs...)
+ode5ms(F, x0, tspan; kwargs...) = ODE.ode_ms(F, x0, tspan, 5; kwargs...)
 
 ###############################################################################
 ## STIFF SOLVERS
@@ -350,7 +350,7 @@ end
 
 #ODEROSENBROCK Solve stiff differential equations, Rosenbrock method
 #   with provided coefficients.
-function oderosenbrock(F, x0, tspan, gamma, a, b, c; jacobian=nothing)
+function oderosenbrock(F, x0, tspan, gamma, a, b, c; jacobian=nothing, kwargs...)
 
     if typeof(jacobian) == Function
         G = jacobian
@@ -403,7 +403,7 @@ const kr4_coefficients = (0.231,
                             6.02015272865   0.1597500684673  0        0
                            -1.856343618677 -8.50538085819   -2.08407513602 0],)
 
-ode4s_kr(F, x0, tspan; jacobian=nothing) = oderosenbrock(F, x0, tspan, kr4_coefficients...; jacobian=jacobian)
+ode4s_kr(F, x0, tspan; jacobian=nothing, kwargs...) = oderosenbrock(F, x0, tspan, kr4_coefficients...; jacobian=jacobian, kwargs...)
 
 # Shampine coefficients
 const s4_coefficients = (0.5,
@@ -417,10 +417,11 @@ const s4_coefficients = (0.5,
                            372/25   12/5    0   0
                           -112/125 -54/125 -2/5 0],)
 
-ode4s_s(F, x0, tspan; jacobian=nothing) = oderosenbrock(F, x0, tspan, s4_coefficients...; jacobian=jacobian)
+ode4s_s(F, x0, tspan; jacobian=nothing, kwargs...) =
+      oderosenbrock(F, x0, tspan, s4_coefficients...; jacobian=jacobian, kwargs...)
 
 # Use Shampine coefficients by default (matching Numerical Recipes)
-ode4s(F, x0, tspan; jacobian=nothing) = ode4s_s(F, x0, tspan; jacobian=nothing)
+ode4s(F, x0, tspan; jacobian=nothing, kwargs...) = ode4s_s(F, x0, tspan; jacobian=nothing, kwargs...)
 
 const ms_coefficients4 = [ 1      0      0     0
                           -1/2    3/2    0     0
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -5,6 +5,9 @@ module ODE
 
 using Polynomials
 using Compat
+using DiffEqBase
+
+import DiffEqBase: solve
 
 ## minimal function export list
 # adaptive non-stiff:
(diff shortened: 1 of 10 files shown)


src/common.jl
```diff
@@ -0,0 +1,99 @@
+abstract ODEJLAlgorithm <: AbstractODEAlgorithm
+immutable ode23Alg <: ODEJLAlgorithm end
```
mauro3: I'd make those `Ode23Alg`, etc.

src/common.jl
```diff
@@ -0,0 +1,99 @@
...
+
+    tspan = prob.tspan
+
+    if tspan[end]-tspan[1]<tType(0)
+        error("final time must be greater than starting time. Aborting.")
+    end
+
+    u0 = prob.u0
+
+    Ts = sort(unique([tspan[1];saveat;tspan[2]]))
```
mauro3: Why do you sort those?

test/common.jl
```diff
@@ -0,0 +1,21 @@
+using ODE, DiffEqBase, DiffEqProblemLibrary
```
mauro3: Can you test something more here than just throws-no-error?  Maybe shape of solution? Within certain tolerance?

mauro3:
Thanks!

kmsquire:
> I had to put the Alg in the type name because otherwise it matches the function's name. If anyone has a better idea of how to handle this let me know.

Aren't functions also types now, as of Julia 0.5?

ChrisRackauckas:
>Aren't functions also types now, as of Julia 0.5?

Ahh yes. I actually see that the problem was that I was putting the type declaration after the function declaration. If you have that the actual `ode23` function adds a dispatch to the already existing `ode23` function (from the constructor), it works:

```julia
immutable ode23 end

ode23(fn, y0, tspan; kwargs...) = 
```

So then the question is just whether you prefer it as `ode23()` to match the function or `Ode23` since it's a type. With LSODA.jl and ODEInterface.jl I chose to match the function names: `dopri5()` and `lsoda()`.

test/common.jl
ChrisRackauckas: Indeed. That's (and missing some algorithms) is why I had it as WIP.

ChrisRackauckas:
This was updated to have the `ode23()` syntax and the tests check a bit more now. Notice that the tests on `ode23s()` were disabled: I think that might be an error in the actual function itself? It's only returning the initial timepoint.

test/common.jl
ChrisRackauckas: Guess I never did name it WIP, my bad. Should be complete now though, sans the ode23s issue which I think is beyond this PR.

src/common.jl
```diff
@@ -80,6 +74,33 @@ function solve{uType,tType,isinplace,algType<:ODEJLAlgorithm,F}(prob::AbstractOD
...
+    elseif typeof(alg) <: ode4s
+        ts,timeseries_tmp = ODE.ode4s(f,u0,Ts,
+                          norm = norm,
+                          abstol=abstol,
+                          reltol=reltol,
+                          maxstep=dtmax,
+                          minstep=dtmin,
+                          initstep=dt,
+                          points=points)
     end
```
kmsquire: Again, if you're willing to give up the type declaration, and since all of the algorithms accept a `kwargs...` parameter, you could simplify the code greatly:

```julia
function solve{uType,tType,isinplace,F}(prob::AbstractODEProblem{uType,tType,isinplace,F}, ode_alg,...* rest of args *)
    < snip >
    ts,timeseries_tmp = ode_alg(f,u0,Ts,
                           norm = norm,
                           abstol=abstol,
                           reltol=reltol,
                           maxstep=dtmax,
                           minstep=dtmin,
                           initstep=dt,
                           points=points)
```

ChrisRackauckas: >Again, if you're willing to give up the type declaration

No, that would be a huge change and would have some big downsides for usage with things like Sundials.jl.

I could also just do the call via `typeof(alg)` since now they are the same name. The problem is that not all of the algorithms do accept `kwargs` parameters (`ode4` is an example which does not). Maybe the better solution to make them all accept it?

kmsquire: I'm curious why removing the type declaration would be a problem for usage with Sundials.jl?  Even if this function doesn't have a type on the parameter, those solvers could be passed in.  Or are you trying to prevent, e.g., a DAE algorithm from being passed into a ODE solver?

Either way, if there are only a few `ode*` algorithms which don't accept `kwargs`, then yes, I think giving all of them `kwargs` (and ignoring thes parameters if necessary) is worthwhile, since it let's you simplify this code considerably.

ChrisRackauckas: Oh I thought you meant type constructor. My bad for the confusion. No, the type declaration ode_alg::T with `T<ODEJLAlgorithm` is necessary for dispatch to work correctly, otherwise it doesn't know which `solve` function to use.

And sure, I'll change the others to use `kwargs`.

ChrisRackauckas:
I did the suggested changes. Please take another look. Thanks!

ChrisRackauckas:
I'll wait for @mauro3 's approval.

ChrisRackauckas:
Bump.

ChrisRackauckas:
I'll give @mauro3 a few more days.

src/ODE.jl
```diff
@@ -16,6 +21,9 @@ export ode23s
 # non-adaptive stiff:
 export ode4s
 
+# Common Interface
+export ODEJLAlgorithm
```
mauro3: `ODEjlAlgorithm` would be easier to read.

src/ODE.jl
```diff
@@ -152,7 +160,7 @@ include("runge_kutta.jl")
 
 # ODE_MS Fixed-step, fixed-order multi-step numerical method
 #   with Adams-Bashforth-Moulton coefficients
-function ode_ms(F, x0, tspan, order::Integer)
+function ode_ms(F, x0, tspan, order::Integer; kwargs...)
```
mauro3: Are these `kwargs...` really necessary?  It just seems like the abstraction is a bit leaky if they now need to accept throw away kwargs.

src/algorithm_types.jl
```diff
@@ -0,0 +1,8 @@
+abstract ODEJLAlgorithm <: AbstractODEAlgorithm
+immutable ode23 <: ODEJLAlgorithm end
```
mauro3: This is a bit magical.  How about a comment like:
```
Making the ODE-solver functions into types lets us dispatch on them.  Used in the DiffEqBase interface.
```

src/common.jl
mauro3: Hmm, I'm not a fan of this: it adds unnecessary complexity.  It's not much complexity but feature creep needs to be avoided actively.  That user could easily have done that himself with one or two lines of code.  Now we need to support X lines of code for this instead.

Also, this is wrong when integration is backward in time.  Ah, I see, that is not allowed.  Why not? ODE.jl solver can do it. (I guess I have my feature-creep priorities elsewhere...)

src/common.jl
```diff
@@ -0,0 +1,62 @@
...
+        f = (t,u) -> vec(prob.f(t,reshape(u,sizeu)))
+    else
+        f = prob.f
+    end
+
+    if uType <: AbstractArray
+        u0 = vec(prob.u0)
+    else
+        u0 = prob.u0
+    end
```
mauro3: To save a few LOCs `u0 = uType <: AbstractArray ? vec(prob.u0) : prob.u0`

src/common.jl
```diff
@@ -0,0 +1,62 @@
...
+
+    ts,timeseries_tmp = AlgType(f,u0,Ts;
```
mauro3: Add a comment along the lines: `Calling the solver`

mauro3:
Thanks Chris!  Have you run your bigger test suites using this?

src/ODE.jl
ChrisRackauckas: These kwargs are necessary for the call

```julia
AlgType(f,u0,Ts;
                      norm = norm,
                      abstol=abstol,
                      reltol=reltol,
                      maxstep=dtmax,
                      minstep=dtmin,
                      initstep=dt,
                      points=points)
```

to not throw an error.

ChrisRackauckas:
>Hmm, I'm not a fan of this: it adds unnecessary complexity. It's not much complexity but feature creep needs to be avoided actively. That user could easily have done that himself with one or two lines of code. Now we need to support X lines of code for this instead.

>Also, this is wrong when integration is backward in time. Ah, I see, that is not allowed. Why not? ODE.jl solver can do it. (I guess I have my feature-creep priorities elsewhere...)

Ahh yes, I forgot ODE.jl can integrate backwards in time. I got rid of the sort and the error for this.

ChrisRackauckas:
I pushed an update which addressed most of what you mentioned. Tests on the nightly will fail because of an unrelated thing (macro hygiene change, which has been fixed, but a tag needs to happen somewhere else in the ecosystem for it to pass).

I got rid of the sorting and time direction check because indeed, one of the nice things about ODE.jl is that it can integrate backwards in time (and I'm hoping to try and get that in other places as well). 

I don't know what you want to do about the `kwargs...` though: they are necessary for this style of implementation, unless you want to `if AlgType <: ...` and pull out each different case. We can subtype the algorithms a bit more and do that, but this is mostly a code "styling" issue so I'll just go with whatever makes others happy here.

>Thanks Chris! Have you run your bigger test suites using this?

Yes, I have already been using it some places. I haven't used it in the full benchmarking suite yet because I am waiting for the PR49 changes to go through so I can update the benchmarks to include all of them. I can't actually run them simultaneously until the package is renamed, so I've just been holding off and doing other things in the meantime.

mauro3:
LGTM

mauro3:
I renamed JuliaODE/ODE.jl to PR49.jl

pwl:
Don't you need to rename the module too?

ChrisRackauckas:
Yes, the module needs to be renamed for it to be referenced correctly.

pwl:
Hm, if we are going to put an effort and rename the whole module anyway, shouldn't we think about a better name then PR49?  We can discussed it in https://github.com/JuliaDiffEq/Roadmap/issues/7.

mauro3:
Sorry, I'll revert it.

tkelman:
You could also dispatch on `typeof(ode45)` and leave them as functions instead of types, now these have zero-argument constructors that don't serve much purpose

ChrisRackauckas:
In other cases they are used as functions in order to specify more type information. Some examples include:

- https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/algorithms.jl#L29
- https://github.com/JuliaDiffEq/Sundials.jl/blob/master/src/algorithms.jl#L8

The ODE.jl algorithms are too simple to need this (at least for right now), but they are setup in a way that is standardized because it would be more difficult to know to use `ode45` sometimes and `Tsit5()` others. It's easier to just always have it have the constructor, which is what JuMP, Optim, and JuliaML do as well (so users should see some cross-Julia continuity in how algorithm choice is done).

PR #122, opened 2017-04-10
Update common interface

ChrisRackauckas:
This updates the common interface to the new specs for DiffEqBase.jl's v0.15.0. It is incompatible with previous versions. This should not be merged until DiffEqBase.jl v0.15.0 is released, and upper bounds are placed in METADATA.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -3,6 +3,7 @@ function solve{uType,tType,isinplace,AlgType<:ODEjlAlgorithm}(prob::AbstractODEP
     save_timeseries=nothing,
     saveat=tType[],reltol = 1e-5, abstol = 1e-8,
     save_everystep=isempty(saveat),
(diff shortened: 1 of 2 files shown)


src/common.jl
```diff
@@ -41,16 +65,26 @@ function solve{uType,tType,isinplace,AlgType<:ODEjlAlgorithm,F}(prob::AbstractOD
                       initstep=dt,
                       points=points)
 
-    # Reshape the result if needed
+
+    if save_start
```
gabrielgellner: Might be worth adding a comment here to explain these constants.

gabrielgellner: Sorry I take it back. the `save_start` -> `start_idx` is actually clear. I just missed this part of the updated api.

gabrielgellner:
LGTM!

PR #131, opened 2017-05-11
Fix deprecation warnings on v0.6

ScottPJones:


ChrisRackauckas:
This only failed on AppVeyor because it ran before the necessary tag.

PR #132, opened 2017-05-11
Fix #166, add checks for keyword arguments to make sure valid for thi…

ScottPJones:
…s solver

### DIFF
--- a/src/ODE.jl
+++ b/src/ODE.jl
@@ -10,7 +10,7 @@ using DiffEqBase
 import DiffEqBase: solve, check_keywords
 
 const warnkeywords =
-    (:save_idxs, :d_discontinuities, :isoutofdomain, :unstable_check,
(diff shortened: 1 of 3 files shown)


src/ODE.jl
```diff
@@ -7,7 +7,16 @@ using Polynomials
 using Compat
 using DiffEqBase
 
-import DiffEqBase: solve
+import DiffEqBase: solve, check_keywords
+
+const warnkeywords =
+    (:save_idxs, :d_discontinuities, :isoutofdomain, :unstable_check,
```
ChrisRackauckas: ODE.jl can do `isoutofdomain`

ScottPJones: Done!

ScottPJones:
I think we need to look at how the coverage is dealt with, because we probably only want to do it once, maybe on latest.

